PR #3, opened 2024-01-09
Add SPI v2 RFC

tdcmeehan:


RFC-0001-native-spi.md
```diff
@@ -0,0 +1,548 @@
...
+
+Functions returned from this endpoint should behave as built-in functions to the user.  Built-in functions are currently simply functions that don’t require a catalog and schema to be specified when they are executed.  Because each `FunctionNamespaceManager` registers itself to the coordinator with a catalog and namespace to refer, we must allow the functions from the `NativeFunctionNamespaceManager` to be addressed as if they were built-in functions.  To support this, the built-in namespace will be become configurable, and a `FunctionNamespaceManager` can indicate to the coordinator that its own namespace should be considered the default namespace.  The coordinator will validate that only one such plugin has specified itself as a default namespace.  The `NativeFunctionNamespaceManager` will indicate that it is a default namespace, and hence will replace built-in functions in Java.
+
+![Function flow](RFC-0002/functions.png)
+
+
+
+#### How will constant folding work
```
yingsu00, line 245: @tdcmeehan Would it be possible to add the explanation what is exactly constant folding here and why it should be treated as normal functions? It'll be nice to give an example.

RFC-0001-native-spi.md
```diff
@@ -0,0 +1,548 @@
...
+
+A new endpoint will be added, `/v1/functions/{function_name}`.  A `GET` to this method will return function output as a page with just one block and just one row.  It will take in parameters `arg1`, `arg2`, …, `argN` and `returnType`, all of which will be serialized `TypeSignature` values.  The C++ sidecar will interpret the type signatures and resolve a function, convert the Presto page, evaluate the function, then return the results as a Presto page and send back to the coordinator.
+
```
yingsu00, line 248: Would it be possible to add your offline explanation about the 2 options and their pros and cons here? 

"from that point is no different from any other function, so I think it makes most sense to do it in the sidecar (alternative is to keep track of what functions we want to duplicate in Java and what functions we don't)"

Also, IF the performance of the simple functions, shipped to the sidecar, is a lot slower than in Java process, what can we do to improve that?

RFC-0001-native-spi.md
```diff
@@ -0,0 +1,574 @@
...
+#### Configuration properties 
+
+Presto validates which configuration properties end up being used by ensuring that all entries in the configuration correspond to a configuration file in the codebase. If one references a configuration that is unused then the Presto server will fail to start. In Java clusters, there are essentially three categories of configuration:
+
+
+
+1. Configuration that affects coordinators only
+2. Configuration that affects workers only
+3. Configuration that affects both
```
yingsu00, line 106: Hi @tdcmeehan would it be possible to add some examples for the 3 categories?

yingsu00 (approved):

RFC-0001-native-spi.md
tdcmeehan: Done

RFC-0001-native-spi.md
tdcmeehan: Done, PTAL!

RFC-0001-native-spi.md
tdcmeehan: Added, PTAL!

PR #5, opened 2024-02-20
Add Iceberg samples

ZacBlanco:


ZacBlanco:
From the original thread from @aaneja:

> Some questions I had -
> What would be the schema for the $samples table ? Would there be any extra columns besides the ones from the underlying table

The schema should exactly map to the real table. Any updates to the real table's schema would also happen on the sample table.

> Would we store some metadata for the $samples table that could be used to identify -
>         What table sampling method was used to seed the sample
>         When was the sample last updated w.r.t to the changelog

This was not mentioned in the RFC, but our prototype implementation included this info. We stored it in the iceberg table properties. I will add this to the RFC

RFC-0001.md
```diff
@@ -0,0 +1,230 @@
...
+`map<string, string>`.
+
+#### Sample Maintenance
+
+Sample maintenance can be performed through a set of SQL statements which replace records in the
+sample with updates, deletes, or inserts that occur on the real table. With the introduction of
+Iceberg's changelog table in #20937 we can efficiently update the sample after the table has been
+updated without needing to re-sample the entire table.
+
+Incremental sample maintenance will be out of scope for the main PR, but all of the infrastructure
```
tdcmeehan, line 161: I think we can do better than this comment.  We can say, with the help of an orchestration tool, sample maintenance using Presto or Spark is as simple as running the following queries on a fixed cadence according to their desire for up to date samples--right?

tdcmeehan: Or, once #12 is merged, perhaps we can add a new distributed procedure which updates the table samples?

hantangwangd: > Or, once #12 is merged, perhaps we can add a new distributed procedure which updates the table samples? 

Sure, in terms of feasibility, I think we can do this through a customized distributed procedure.

RFC-0001.md
```diff
@@ -0,0 +1,230 @@
...
+for sample maintenance is available already once the samples are generated. The biggest hurdle for
+users will be that sample maintenance will need to be done manually. We can document the maintenance
+process for samples, but there currently isn't a way to schedule or automatically run the correct
+set of maintenance queries for the sample tables.
```
hantangwangd, line 165: Besides of incremental sample maintenance, should we describe the mechanism of sample table refreshment as well? How to refresh the sample table and what granularity of refreshment is supported.

In addition, should we maintain the snapshot version correspondence between the sample table and the source table? As we do analyzing on the sample table, meanwhile do querying on the source table. We need a way to find the nearest statistics on sample table for the specified snapshot version on source table.

RFC-0001.md
```diff
@@ -0,0 +1,230 @@
...
+As long as the metadata stored is limited to something small, say, less than 1KiB, then it is
+probably fine to store this information inside of the iceberg's table properties which is a simple
+`map<string, string>`.
+
+#### Sample Maintenance
+
+Sample maintenance can be performed through a set of SQL statements which replace records in the
+sample with updates, deletes, or inserts that occur on the real table. With the introduction of
+Iceberg's changelog table in #20937 we can efficiently update the sample after the table has been
+updated without needing to re-sample the entire table.
```
hantangwangd, line 159: Iceberg changelog  table currently do not support V2 row level deletion. It seems to be a big problem for incremental maintenance.

RFC-0001.md
```diff
@@ -0,0 +1,230 @@
...
+We should be able to see improvement in query plans and execution times without running full
+table scans to generate the samples. Additionally, cost of maintaining statistics should decrease.
+I would recommend measuring the amount of data scanned across multiple ANALYZE queries and compare
+it to the amount of data scanned for sample creation and maintenance.
+
+## [Optional] Other Approaches Considered
+
+### Table-Level Storage in Puffin files
+
+One alternative approach is to generate the samples and store them in Puffin files[^puffin_files].
```
aditi-pandit, line 184: Its better to write a technical doc in a third person tone.

aditi-pandit (commented):
Thanks @ZacBlanco for this doc. Gives a good picture of the work being done.

PR #7, opened 2024-03-13
RFC for Global Row IDs

elharo:
@tdcmeehan

### DIFF
--- a/RFC-0001-row-id.md
+++ b/RFC-0001-row-id.md
@@ -28,13 +28,13 @@ partition is mutated.
 ## Background
 
 Queries can produce lists of row IDs that can be saved and used as input to
-future queries for quick selection. A list of row IDs can be used include or
-exclude rows to form a subset from a table.
+future queries for quick selection. A list of row IDs can be used to include or
+exclude rows in way that identifies a subset of a table.
 
 ### Goals
 
 Row IDs resolve to the same value across compute engines. In this way they can
-be used as a strong-equality comparison across compute engines.
+be used for strong-equality comparison across compute engines.
 
 ### Non-goals
 
@@ -53,11 +53,12 @@ The exact format of the row ID is connector specific and not constrained by this
 RFC. It might change from one warehouse to the next if different warehouses use
 different metatdata servers to supply the partition half of a row ID.
 
-The metadata server (e.g. Hive Metastore) supplies the partition component in
+The metadata server (for example, Hive Metastore) supplies the partition component in
 the same way it supplies all the other information in the `Partition` object.
 Typically this involves a network call to the server which returns a Thrift
 object. The fields of the Thrift object are deserialized and loaded into a
-`Partition` object.
+`Partition` object. One of these fields is a byte array containing the
+"row ID partition component".
 
 When a connector reads rows from a file, it passes this partition component to
 the reader. Assuming the SQL query explicitly references the `$row_id`
@@ -71,8 +72,8 @@ the reader can use to construct a row ID from the row number and the row group.
 ## Adoption Plan
 
 This should be completely ignorable for all existing clients that use Presto.
-Everything is API compatible, and any SQL queries that don't access a column
-named `$row_id` will be unaffected.
+Everything is API compatible, and no table or SQL query should have a column named
+`$row_id` (or anything else that begins with a dollar sign).
 
 ## Test Plan
 
--- a/RFC-0001-row-id.md
+++ b/RFC-0001-row-id.md
@@ -1,6 +1,6 @@
 # **RFC 1 for Presto**
 
-## Row ID
+## Row ID for Hive
 
 Proposers
 
@@ -44,8 +44,9 @@ Comparison and sorting of row IDs.
 
 A row ID is a VARBINARY pseudo-column, also known as a hidden column. Each row
 will have a unique value that can be compared for equality, byte by byte. The
-type is not orderable. In SQL queries this will have the name `$row_id` and the
-type ROW_ID, a new built-in type. The complete row ID will have two components,
+type is not orderable. In SQL queries this will have the name `$row_id` and
+the type `VARBINARY`. In Java it will have the type `SQLVarBinary`.
+The complete row ID will have two components,
 the first to identify a row within a partition and the second to identify the
 partition within the warehouse.
 
@@ -53,27 +54,27 @@ The exact format of the row ID is connector specific and not constrained by this
 RFC. It might change from one warehouse to the next if different warehouses use
 different metatdata servers to supply the partition half of a row ID.
 
-The metadata server (for example, Hive Metastore) supplies the partition component in
-the same way it supplies all the other information in the `Partition` object.
+The metadata server (for example, Hive Metastore) supplies the partition component
+in the same way it supplies all the other information in the `Partition` object.
 Typically this involves a network call to the server which returns a Thrift
 object. The fields of the Thrift object are deserialized and loaded into a
 `Partition` object. One of these fields is a byte array containing the
-"row ID partition component".
+"row ID partition component". This byte array will be passed down the chain into a
+`HiveSplit` and eventually into a `PageSource`.
 
-When a connector reads rows from a file, it passes this partition component to
-the reader. Assuming the SQL query explicitly references the `$row_id`
-pseudo-column, the reader constructs a unique row component for each row from
-the row number and the row group. It concatenates this row component with the
-partition component and returns the combined value as the `$row_id` field.
-Possibly instead of passing the partition component directly, it passes a function
-the reader can use to construct a row ID from the row number and the row group.
-
+If the SQL query explicitly references the `$row_id`
+pseudo-column, when a page source reads rows from a file, it sets `appendRowNumber` to true 
+so it receives row numbers from the reader. Then the page source coerces the 
+row number block (type `long`) into a row ID block (type `SqlVarBinary`) by concatenating
+the bytes of the row number, the row group ID, and the partition ID into a single 
+blob.
 
 ## Adoption Plan
 
 This should be completely ignorable for all existing clients that use Presto.
 Everything is API compatible, and no table or SQL query should have a column named
-`$row_id` (or anything else that begins with a dollar sign).
+`$row_id` (or anything else that begins with a dollar sign that is not a documented
+Presto pseudo-column).
 
 ## Test Plan
 
--- a/RFC-0001-row-id.md
+++ b/RFC-0001-row-id.md
@@ -1,4 +1,4 @@
-# **RFC 1 for Presto**
+# **RFC 2 for Presto**
 
 ## Row ID for Hive
 
--- a/RFC-0001-row-id.md
+++ b/RFC-0001-row-id.md
@@ -1,6 +1,6 @@
 # **RFC 2 for Presto**
 
-## Row ID for Hive
+## Hive Row IDs
 
 Proposers
 
--- a/RFC-0001-row-id.md
+++ b/RFC-0001-row-id.md
@@ -1,0 +1,80 @@
+# **RFC 1 for Presto**
+
+## Row ID
+
+Proposers
+
+* Brian Dahmen
+* Devang Shah
+* Elliotte Rusty Harold
+* Serge Druzkin
+* Sreeni Viswanadha
+
+## Related Issues
+
+* Row ID support #22078
+* Binary IDs for Hive partitions #22103
+* Binary rowID type #22053
+* Start adding row\_id hidden column to Hive #22008
+
+
+## Summary
+
+A row ID is a binary key for a row that is globally unique across the warehouse.
+It identifies a specific row at a specific version of a partition’s data. A row
+ID will continue to locate the same row in the warehouse until the underlying
+partition is mutated.
+
+## Background
+
+Queries can produce lists of row IDs that can be saved and used as input to
+future queries for quick selection. A list of row IDs can be used include or
+exclude rows to form a subset from a table.
+
+### Goals
+
+Row IDs resolve to the same value across compute engines. In this way they can
+be used as a strong-equality comparison across compute engines.
+
+### Non-goals
+
+Comparison and sorting of row IDs.
+
+## Proposed Implementation
+
+A row ID is a VARBINARY pseudo-column, also known as a hidden column. Each row
+will have a unique value that can be compared for equality, byte by byte. The
+type is not orderable. In SQL queries this will have the name `$row_id` and the
+type ROW_ID, a new built-in type. The complete row ID will have two components,
+the first to identify a row within a partition and the second to identify the
+partition within the warehouse.
+
+The exact format of the row ID is connector specific and not constrained by this
+RFC. It might change from one warehouse to the next if different warehouses use
+different metatdata servers to supply the partition half of a row ID.
+
+The metadata server (e.g. Hive Metastore) supplies the partition component in
+the same way it supplies all the other information in the `Partition` object.
+Typically this involves a network call to the server which returns a Thrift
+object. The fields of the Thrift object are deserialized and loaded into a
+`Partition` object.
+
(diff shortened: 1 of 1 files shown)


RFC-0001-row-id.md
```diff
@@ -0,0 +1,80 @@
...
+
+* Row ID support #22078
+* Binary IDs for Hive partitions #22103
+* Binary rowID type #22053
+* Start adding row\_id hidden column to Hive #22008
+
+
+## Summary
+
+A row ID is a binary key for a row that is globally unique across the warehouse.
```
tdcmeehan, line 23: By "warehouse", do we mean data lake?

RFC-0001-row-id.md
```diff
@@ -0,0 +1,80 @@
...
+It identifies a specific row at a specific version of a partition’s data. A row
+ID will continue to locate the same row in the warehouse until the underlying
+partition is mutated.
```
tdcmeehan, line 26: Suppose I am using open source Hive, but not with ACID transaction support.  How does this look, and is it supported?

RFC-0001-row-id.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Background
+
+Queries can produce lists of row IDs that can be saved and used as input to
```
tdcmeehan, line 30: If they are saved, and the underlying partitions are not sealed, how is it safe to save it?  I.e. there could be extra data which has been ingested or deleted?

RFC-0001-row-id.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Related Issues
+
+* Row ID support #22078
+* Binary IDs for Hive partitions #22103
+* Binary rowID type #22053
+* Start adding row\_id hidden column to Hive #22008
+
+
+## Summary
```
tdcmeehan, line 21: I would like to see more discussion on what part of this falls on the connector and what part falls on the engine.  As I'm reading this, it seems all of these concepts can be scoped purely to the connector--is that right?

RFC-0001-row-id.md
```diff
@@ -0,0 +1,80 @@
...
+A row ID is a binary key for a row that is globally unique across the warehouse.
+It identifies a specific row at a specific version of a partition’s data. A row
+ID will continue to locate the same row in the warehouse until the underlying
+partition is mutated.
+
+## Background
+
+Queries can produce lists of row IDs that can be saved and used as input to
+future queries for quick selection. A list of row IDs can be used include or
+exclude rows to form a subset from a table.
```
tdcmeehan, line 32: To exclude rows, you also need some sort of consistency, right?  i.e. what would happen if more data is added to a partition, but we've saved the row ids and are using it to form that subset?  It seems this proposal is missing that component?

RFC-0001-row-id.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Goals
+
+Row IDs resolve to the same value across compute engines. In this way they can
```
tdcmeehan, line 36: It seems this goal is something a table format defines, not Presto or any other compute engine.  Hive defines the behavior for ACID transactions, which includes row IDs.  Iceberg similarly defines positions, which I think is a transferable concept.  Hudi uses primary keys, which I don't think is transferable.  My concern is we're defining concepts in the wrong place--storage defines the behavior in all the examples I listed, and Presto conforms to the spec they give.  If Presto defines it, basically it becomes esoteric Presto behavior, it's not really a spec at all.

If this is originating from Metalake, shouldn't this be something that is defined internally by Metalake?  In which case, don't you already have connectors and infrastructure that speak in terms of "metalake"?  What aspect of this proposal necessitates any change at all in core, given we already have fully functional examples of this concept purely at the connector level?

RFC-0001-row-id.md
```diff
@@ -0,0 +1,80 @@
...
+be used as a strong-equality comparison across compute engines.
+
+### Non-goals
+
+Comparison and sorting of row IDs.
+
+## Proposed Implementation
+
+A row ID is a VARBINARY pseudo-column, also known as a hidden column. Each row
```
tdcmeehan, line 45: Why does it need to be exposed as a column, instead of being encapsulated by the connector?

RFC-0001-row-id.md
```diff
@@ -0,0 +1,80 @@
...
+will have a unique value that can be compared for equality, byte by byte. The
+type is not orderable. In SQL queries this will have the name `$row_id` and the
```
tdcmeehan, line 47: Wouldn't you want it to be orderable, for efficient pruning of ORC stripes and Parquet blocks?  For example, the Iceberg spec explicitly orders positions for delete files https://iceberg.apache.org/spec/#position-delete-files

RFC-0001-row-id.md
```diff
@@ -0,0 +1,80 @@
...
+type ROW_ID, a new built-in type. The complete row ID will have two components,
+the first to identify a row within a partition and the second to identify the
+partition within the warehouse.
+
+The exact format of the row ID is connector specific and not constrained by this
+RFC. It might change from one warehouse to the next if different warehouses use
+different metatdata servers to supply the partition half of a row ID.
+
+The metadata server (e.g. Hive Metastore) supplies the partition component in
```
tdcmeehan, line 56: Not everything uses a Hive metastore for its catalog.  Does this mean this proposal is specific to the Hive connector?

RFC-0001-row-id.md
```diff
@@ -0,0 +1,80 @@
...
+the same way it supplies all the other information in the `Partition` object.
+Typically this involves a network call to the server which returns a Thrift
+object. The fields of the Thrift object are deserialized and loaded into a
+`Partition` object.
+
+When a connector reads rows from a file, it passes this partition component to
```
tdcmeehan, line 62: I think what's missing from this proposal is why go through the trouble of exposing it as a column--if it's used for "exclud[ing] rows to form a subset from a table", why not do that purely within the connector?

tdcmeehan (changes requested):
Thanks for writing this up!  I've left my questions below.

RFC-0001-row-id.md
elharo: I am assuming partitions are immutable. But even if they're not, a list of row IDs still defines a subset of a table.

RFC-0001-row-id.md
elharo: It's not just the table format. The table format can define a row ID for rows within that table. ORC already does. However, for a global row ID across the warehouse, we need components that identify not just the row within the table but also the table (or rather the version of the partition within the table). That piece is supplied with Metastore.

I'm going to ask @kaikalur and @bmdahmen, to chime in on why this belongs in core.

RFC-0001-row-id.md
elharo: In a SQL query it needs to appear as a column. I think you're only asking about what happens after the query `SELECT $row_id from table_name` begins moving through Presto. I'm going to defer to others to argue for this, but I seem to recall someone mentioning that making this a low level column improves the optimization and query planning. Meta-folk, am I remembering that right?

RFC-0001-row-id.md
elharo: Maybe. I don't think we've ruled it out, and it doesn't look hard to add.

RFC-0001-row-id.md
elharo: No, that's why "e.g." instead of "i.e.".  I'll spell it out. Good tech writing doesn't use Latin. :-)

RFC-0001-row-id.md
elharo: We want to get a list of row IDs, save them somewhere, and then use that list in a completely different query engine like Spark that's operating on the same data files.

Maybe "orm a subset" is a confusing here. I don't mean literally materialize a table or even create a view of it, really. I'm just trying to say that having the list of IDs allows us to conceptually have different subsets of the rows of a table. I'm not sure how to write that more clearly, but let me try again. Maybe "A list of row IDs can be used to include or exclude rows in way that identifies a subset of a table."

RFC-0001-row-id.md
tdcmeehan: Right, but the table, the version of the table, are both things that also are connector-specific.  They may be entirely encapsulated in the connector.

RFC-0001-row-id.md
tdcmeehan: OK, this doc is missing, what use does one have for $row_id, and how is it used for query optimization?  Why can't $row_id be connector-supplied?

RFC-0001-row-id.md
tdcmeehan: I find the definition of [data lake](https://en.wikipedia.org/wiki/Data_lake) on Wikipedia more relevant for this proposal than [data warehouse](https://en.wikipedia.org/wiki/Data_warehouse).

RFC-0001-row-id.md
tdcmeehan: I agree this is safe to do in your environment, but what I'm wondering is, because many of these changes will happen in Hive, is it safe to do in Hive?

Query A is writing data to partitions, it takes an hour to complete.
Query B is identifying the deleted rows and putting them in a file.  It starts at the half hour mark and takes 5 minutes.
We are left with query B's files that only contain partial rows.  Now, I want to identify the rows that query B missed because of the concurrent update--how do I do that?

Because this doesn't seem safe, I am wondering if this can be encapsulated purely within an internal connector that speaks in terms of your warehouse, rather than being added to Hive.

RFC-0001-row-id.md
tdcmeehan: What is the purpose of saving them through an externally provided row id?  Is it to identify removed rows?  If so, why not follow the approach of other connectors and instead just internally encapsulate that as a delete?

RFC-0001-row-id.md
elharo: Per wikipedia "A data lake is a system or [repository of data](https://en.wikipedia.org/wiki/Data_repository) stored in its natural/raw format,[[1]](https://en.wikipedia.org/wiki/Data_lake#cite_note-1) usually object [blobs](https://en.wikipedia.org/wiki/Binary_large_object) or files. A data lake is usually a single store of data including raw copies of source system data, sensor data, social data etc.,[[2]](https://en.wikipedia.org/wiki/Data_lake#cite_note-2)"

Since that includes a lot of non-tabular data, I don't think it describes what Presto operates on.

RFC-0001-row-id.md
elharo: Let me ask you two questions to see if I can understand your points better:

1. If it's all in the connector, then does `SELECT $row_id FROM table_name` still work? That is, can we still expose the $row_id pseudo-column?
2. Can the optimizer still optimize queries that use $row_id?

I think the answer to both questions is yes, but I want to make sure.

tdcmeehan: 1. Yes, `SELECT $row_id FROM table_name` will still work.  Note that `$row_id` is defined within the connector, and uses a type defined by the connector.  That is, there are no code additions in `presto-main`, `presto-common`, or `presto-spi` to facilitate this column name and type.
2. Yes, the optimizer can still optimize queries that use `$row_id`.  These columns can be identified by the connector as being unique identifiers, and as such the constraints framework may optimize them, the same as other columns that contain unique values. https://github.com/prestodb/presto/issues/16413. I think it's important that the way the optimizations are applied be general, as for example I noted below, the $row_id concept doesn't seem applicable to Hudi, whereas constraints are applicable to Hudi, Iceberg positions, the $row_id concept mentioned here, and are broadly applicable to any unique column for any data sources.

tdcmeehan: I'm wondering if we should split this proposal into two parts:

1) Exposing $row_id in a Hive-specific, not general way.  While I worry that exposing this without guaranteeing ACID constraints is a questionable decision, I think we could make progress there (e.g., by somehow making sure it can't be persisted, or by making whatever minimal plumbing changes are required in the Hive connector, but only PrismConnector (internal) exposes it).  
2) Talking about how to get the optimizer to leverage unique columns in general, and debating the best path forward for that.

I think if the proposal and code changes were limited to 1, we could make quick progress.  I just worry 2 requires a deeper design, and we really need to settle on how to do that, as it seems we're thinking of two parallel directions there, and we shouldn't put out code that does this two different ways.

RFC-0001-row-id.md
tdcmeehan: Presto not operating over data lakes is news to lots of people. :)  I don't mean to be nit-picky, but I think outside of Meta, when you use the term Warehouse, it typically means Enterprise Data Warehouse--which are data appliances that include a front (engine) and back (storage) component.  Presto can operate over that, but this proposal talks about identifying row IDs over files and partitions on distributed storage that is cataloged by an HMS or similar system, which is a lake.  Lakes often include unstructured data in addition to structured files, and there are other systems that use this parallel to Presto.  This flexibility in the structure of the data is what makes them appealing, and doesn't have any bearing on Presto's ability over to operate over the lake.

RFC-0001-row-id.md
elharo: You've been working on this for longer than I have so I suspect you're seeing some approaches I'm just not thinking of. When you say encapsulate it entirely in the connector, which classes specifically are you suggesting this goes into?

By comparison, what I'm currently thinking is that Metastore stuffs the partition ID into HiveSplit. (This is roughly #22103). That HiveSplit eventually arrives as an argument in HivePageSourceProvider.createSelectivePageSource.createAggregatedPageSource
 
At this point we check to see if the split has a partition ID. If it does, we pass that partition ID down the call chain until it arrives in OrcSelectiveRecordReader.getNextPage. (code not yet written)

Now here there is already code that appends a block of row numbers if the right flag is set. I'm thinking that if a partition ID is present and if $row_id is a requested column, then here we add the block of row IDs by concatenating the row number and partition ID. I also need to stuff the row group ID in here somehow. Then it all moves back up the call stack.

There might well be some other place we can do this, and I'm just not seeing it. Can you elaborate on the alternative you'd prefer? Thanks.

tdcmeehan: The flow you outline is all connector-specific: the Metastore is a Hive concept, as is the ORC reader (although other connectors use that part a library).

There's two parts that I'd like to think about:

1) Can the `$row_id` hidden column be added to `PrismConnector`?  If this is really hard, could we at least configure this so it's not exposed by default?  It revolves around what purpose there is around exposing row ids when you could have phantom rows.
2) Can the binary `RowIdType` itself be exposed as a new type in `PrismConnector`?  This would be a change in Prism's implementation of `Plugin#getTypes`, instead of a new type in `presto-common`.  This would encapsulate the type inside of where it's used, in `PrismConnector`.

elharo: Looks like PrismConnector went away about 6 years ago. Do you mean HiveConnector?

tdcmeehan: It should be subclass of the Hive connector you currently use.

kaikalur:
I'm not sure why we are syaing no to rowid type. it will be quite useful for optimizations.

RFC-0001-row-id.md
```diff
@@ -0,0 +1,82 @@
...
+
+A row ID is a VARBINARY pseudo-column, also known as a hidden column. Each row
+will have a unique value that can be compared for equality, byte by byte. The
+type is not orderable. In SQL queries this will have the name `$row_id` and
+the type `VARBINARY`. In Java it will have the type `SQLVarBinary`.
+The complete row ID will have two components,
+the first to identify a row within a partition and the second to identify the
+partition within the warehouse.
+
+The exact format of the row ID is connector specific and not constrained by this
```
tdcmeehan, line 53: Is this RFC now Hive-connector specific?

tdcmeehan: I think this proposal is now a more localized change to the Hive connector (although it is not reflected in this text).  If that's the case, I think we can could close this RFC.

elharo:
OK. I'll keep the branch until I can figure out where to put the docs.

PR #8, opened 2024-03-13
Remove numbering from RFC name

rschlussel:
I'm proposing that we remove the numbering from RFC file names, as I'm not sure it provides a lot of benefit, but it does produce confusion. Currently there are 5 open PRs for RFCs and all are called RFC-0001-<feature-name>.  I'm about to open a 6th and have no idea what number to use.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ RFCs are located in their own repository.
 To create one:
 
 1. Fork the https://github.com/prestodb/rfcs repository
-2. Copy the template file `RFC-0000-template.md` to `RFC-00xx-your-feature.md` and fill it out with your proposal.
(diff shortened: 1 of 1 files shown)


tdcmeehan (approved):
SGTM.  I think the idea is when you go to merge it, you change the number, but I think the actual number is less important, as we have the link in code to reference.

tdcmeehan:
One idea that would be neat: if we keep the numbering, then we could configure an autolink to an RFC (see: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-autolinks-to-reference-external-resources).  For example, RFC-0002 would autolink to https://github.com/prestodb/rfcs/RFC-0002.md (in the same way numbers autolink to issues and PRs).

rschlussel:
that's neat! i'll close this for now.

tdcmeehan:
I've implemented it, so let's look forward to RFC-0001.

PR #10, opened 2024-04-27
Add RFC for Arrow Flight connector

elbinpallimalilibm:


RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### Query execution using flight
+
+![Query execution using Flight](arrow-flight-connector/Query-execution-using-flight.png)
+
+### Low level design
+
+#### Connector splits in Presto for FlightInfo tickets
+
+Flight server can serve data using multiple tickets based on the connector type and query. Data can be fetched in parallel using the multiple flight tickets. Presto can create connector splits for each Flight Info ticket so that data can be fetched in parallel by different worker nodes.
```
tdcmeehan, line 52: Perhaps add details on how FlightInfo is queried?  How do we address that this accepts an opaque blob which may differ from server to server?

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,70 @@
...
+
+![Connector splits in Presto](arrow-flight-connector/Connector-splits-in-Presto.png)
+
+#### Class diagram
```
tdcmeehan, line 56: ArrowQueryBuilder seems to reference SQL.  Why is this?  Is there any presumption regarding the format of the Arrow server in this template?

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,70 @@
...
+
+Create a base Arrow Flight module in Presto which can be extended to create an Arrow Flight connector.
+
+## Background
+
+Apache Arrow Flight is a new general-purpose client-server framework that simplifies high performance transport of large datasets over network interfaces.
+
+Presto currently supports different datasources through connectors for each type of datasource. The connectors could be using JDBC or some other protocol for datasource connection. If we use Flight client to connect to the different datasources we can improve the performance of the datasource connection. Also we would only need to develop one Flight client that can be used across different datasources instead of developing individual connectors for each type of datasource.
+
+## Proposed Implementation
```
tdcmeehan, line 21: Should we also include the plan for Prestissimo/C++?

tdcmeehan (commented):
Thank you for the diagrams, this really helps!

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Summary
+
+Create a base Arrow Flight module in Presto which can be extended to create an Arrow Flight connector.
+
+## Background
+
+Apache Arrow Flight is a new general-purpose client-server framework that simplifies high performance transport of large datasets over network interfaces.
+
+Presto currently supports different datasources through connectors for each type of datasource. The connectors could be using JDBC or some other protocol for datasource connection. If we use Flight client to connect to the different datasources we can improve the performance of the datasource connection. Also we would only need to develop one Flight client that can be used across different datasources instead of developing individual connectors for each type of datasource.
```
jaystarshot, line 19: Are there any preliminary estimates of the performance improvement? What exactly does "performance of the datasource connection" refer to?

BryanCutler:
Thanks for the proposal @elbinpallimalilibm , I think it sound like a useful addition. You hit on the main benefits already, with Flight being a standard RPC protocol, that would open up Presto to connect with any data source that is reachable by a Flight endpoint. Performance-wise, there can definitely be benefits, such as connecting to a PostgreSQL Flight endpoint where data is transferred in Arrow format. However, there are also cases where there might not be any benefit, e.g. connecting with a JDBC-based data source there will be overhead converting rows to columns.

Some questions:
1. Will this include other SQL operations besides reading data, such as update, delete, etc?
2. I think there needs to be more details about the FlightDescriptor. I see you mention JSON, however that puts a restriction on the Flight server to accept that format. There is also a lot that would need to be defined in order to use that command to implement the Presto connector interfaces, for example how will list tables?
3. Have you considered using a FlightSQL client? The FlightSQL protocol defines a way for a client to interact with SQL-based data sources, so this provides a standard with commands to list tables, etc. already baked in. The downside is the Flight server would need to implement FlightSQL, not just core Flight.

lidavidm:
Ah, so (1) this is for Presto consuming data from other sources, not Presto serving Arrow data over Flight and (2) this is for generic Flight and not Flight SQL? 

Since, I believe services advertising Flight support are generally doing so via Flight SQL (examples: InfluxDB, Dremio, and the [PostgreSQL adapter](https://github.com/apache/arrow-flight-sql-postgresql)). Flight SQL also is an actual service definition vs Flight which is more a grab bag of patterns and suggestions for building a service.

elbinpallimalilibm:
> Thanks for the proposal @elbinpallimalilibm , I think it sound like a useful addition. You hit on the main benefits already, with Flight being a standard RPC protocol, that would open up Presto to connect with any data source that is reachable by a Flight endpoint. Performance-wise, there can definitely be benefits, such as connecting to a PostgreSQL Flight endpoint where data is transferred in Arrow format. However, there are also cases where there might not be any benefit, e.g. connecting with a JDBC-based data source there will be overhead converting rows to columns.
> 
> Some questions:
> 
> 1. Will this include other SQL operations besides reading data, such as update, delete, etc?
> 2. I think there needs to be more details about the FlightDescriptor. I see you mention JSON, however that puts a restriction on the Flight server to accept that format. There is also a lot that would need to be defined in order to use that command to implement the Presto connector interfaces, for example how will list tables?
> 3. Have you considered using a FlightSQL client? The FlightSQL protocol defines a way for a client to interact with SQL-based data sources, so this provides a standard with commands to list tables, etc. already baked in. The downside is the Flight server would need to implement FlightSQL, not just core Flight.

1. This proposal is only for reading data. Insert, update, delete are not supported.

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: We have tested with a Flight server implementation that accepts SQL statements for read operations. That is why we have a reference to query builder.

lidavidm:
Ok, that sounds reasonable to me. (...I'd like to get away from Flight itself being such a heavy wrapper over things, but that's a long ways away.)

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,70 @@
...
+
+Proposers
+
+* Elbin Pallimalil
+
+## [Related Issues]
+
+## Summary
+
+Create a base Arrow Flight module in Presto which can be extended to create an Arrow Flight connector.
```
elharo, line 13: can you add links to this?

RFC-0003-arrow-flight-connector.md
elharo, line 19: The connectors could be using --> Connectors can use

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,70 @@
...
+
+A template for Flight connector will be created as `presto-base-arrow-flight` module. Since each implementation of flight server will be different, this base module can be extended to create a flight connector that supports a particular implementation. The template will support only read operations.
+
+### High level design
+
+#### Proposed design
+
+![Design diagram](arrow-flight-connector/Presto-flight-client.drawio.png)
+
+The Arrow Flight libraries provide a development framework for implementing a service that can discover metadata and send and receive data streams. Flight APIs are used to retrieve the data from the corresponding table selected. The steps below need to be followed.
```
elharo, line 31: corresponding table selected --> selected table

elharo (commented):
High level meta comment: we'd like to move some of the existing connectors out of the main repo, and in general make builds more modular. This might make a good test project for that goal. Could this connector be implemented in a completely separate repository? If not, why not?

Second meta comment, especially if this connector does need to belong to this repo: presto has noticeable problems with connectors that were written and then abandoned by their initial implementers. How much support does this have by whom? How long can we count on that support?

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: Which links do you want here?

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: @tdcmeehan The proposed implementation is such that only read operations are supported. And for read operations we are building the query and sending this along with table and schema details to the concrete implementation to create a FlightDescriptor. Do you want us to remove any reference to the query from this connector template?

tdcmeehan:
@elharo

> High level meta comment: we'd like to move some of the existing connectors out of the main repo, and in general make builds more modular. This might make a good test project for that goal. Could this connector be implemented in a completely separate repository? If not, why not?

Just for your information, this is something we've discussed before and there's already alignment on pursuing this.  I think we just need someone to spend the effort to move plugins out of the main repo and setup build pipelines for each.  That being said, I don't think it's necessary or advisable to couple these two initiatives, as I think this effort deserves its own focus.

zeroshade:
> 1. This proposal is only for reading data. Insert, update, delete are not supported.

Just curious, is the limitation here that Presto's "connectors" are specifically for Presto connecting to other data sources rather than for consumers interacting with Presto?

Would it make sense to have a separate proposal to add Arrow FlightSQL as a way for users to interact with Presto itself and retrieve column-oriented data to avoid the column->row overhead? (Correct me if i'm wrong, Presto is internally columnar for execution, right?)

tdcmeehan:
@zeroshade as you mentioned, I think this proposal is for Presto itself to interact with Arrow-like systems.  In the future, we could implement support for updates and deletes on top of this proposal.

A separate thing altogether is Arrow FlightSQL support to supplant, or replace, [Presto's client protocol](https://prestodb.io/docs/current/develop/client-protocol.html).  This has been discussed before (see https://github.com/prestodb/presto/issues/19419), and I believe there's alignment to support that, it just needs someone to pick up this work and come up with a comprehensive design.  That ought to be separate RFC.

What this client-oriented RFC would buy us is, client's wouldn't have to go through the inconvenience of translating row-oriented JSON into Arrow buffers, however I don't think there would be a major performance improvement by doing this, because Presto's client architecture is already bottlenecked by the a single coordinator process, which limits parallelism and makes Presto not very suitable for very large reads.  A larger effort might entail perhaps moving fetching from the coordinator to workers, to give the client greater flexibility in the level of parallelism on fetching the data.  This effort of increasing parallelism for the reads [has long been planned](https://github.com/prestodb/presto/wiki/Presto-v2-client-protocol).  As part of an effort to move toward Arrow Flight, this might entail exposing Tickets for each worker.

I hope this context is helpful.

elbinpallimalilibm:
> High level meta comment: we'd like to move some of the existing connectors out of the main repo, and in general make builds more modular. This might make a good test project for that goal. Could this connector be implemented in a completely separate repository? If not, why not?
> 
> Second meta comment, especially if this connector does need to belong to this repo: presto has noticeable problems with connectors that were contributed and then abandoned by their initial implementers. How much support does this have by whom? How long can we count on that support?

IBM will continue to support and maintain the Arrow Flight connector.

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: We have an initial version of Arrow Flight connector running. Response time for SELECT query seem to be about 30 to 60 percent faster than JDBC, depending on the number of rows in the table.

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,70 @@
...
+1. Create a Flight Client using the host, port, ssl information of the Flight service where the route is opened. 
+2. Use the Flight Client to authenticate to the Flight service. 
+3. Create a FlightDescriptor using a JSON document containing the description of the connection to read or write. The description consists of: 
+    1. Database name, host, port, username and password 
+    2. Interaction properties including schema and table selected or the select statement 
+4. Use the FlightDescriptor to obtain a FlightInfo object. 
+5. Use the FlightInfo object to obtain the list of endpoints we will be retrieving data from. 
+6. Use each endpoint to obtain a Ticket. 
+7. Use the Ticket to obtain a Stream. 
+8. Use the Stream to read data.
```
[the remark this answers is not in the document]
elbinpallimalilibm: Added. details below.

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: Added the details in this section

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,83 @@
...
+### High level design
+
+#### Proposed design
+
+![Design diagram](arrow-flight-connector/Presto-flight-client.drawio.png)
+
+The Arrow Flight libraries provide a development framework for implementing a service that can discover metadata and send and receive data streams. Flight APIs are used to retrieve the data from the selected table. The steps below need to be followed. 
+
+1. Create a Flight Client using the host, port, ssl information of the Flight service where the route is opened. 
+2. Use the Flight Client to authenticate to the Flight service.
```
tdcmeehan, line 36: Can you add more details on how authentication will work?  What fields from the session will be used for authentication?

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,83 @@
...
+
+### High level design
+
+#### Proposed design
+
+![Design diagram](arrow-flight-connector/Presto-flight-client.drawio.png)
+
+The Arrow Flight libraries provide a development framework for implementing a service that can discover metadata and send and receive data streams. Flight APIs are used to retrieve the data from the selected table. The steps below need to be followed. 
+
+1. Create a Flight Client using the host, port, ssl information of the Flight service where the route is opened.
```
BryanCutler, line 35: Does this mean that each task will be creating a Flight client? There is some cost for doing that, I believe mostly the gRPC channel, so it might be worth looking into if there is a benefit to reuse or share clients rather than creating many new ones. @lidavidm might have some insight on this.

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,83 @@
...
+    1. The concrete implementation of this connector template will define a class that returns a byte array for the `FlightDescriptor.command` method. 
+    2. The concrete implementation will specify the necessary details for the database connection in the `command` byte array.
+4. Use the FlightDescriptor to obtain a FlightInfo object. 
+5. Use the FlightInfo object to obtain the list of endpoints we will be retrieving data from. 
+6. Use each endpoint to obtain a Ticket. 
+7. Use the Ticket to obtain a Stream. 
+8. Use the Stream to read data. 
+    1. Get VectorSchemaRoot from Flight stream
+    2. For each column in the result, build a Block from the FieldVector
+    3. Return a page with all the constructed Blocks.
```
BryanCutler, line 48: So a page will roughly correspond to an Arrow RecordBatch. Is there some configuration planned to set the RecordBatch size or will it be specific for the connector/datasource being used?

RFC-0003-arrow-flight-connector.md
lidavidm: Each gRPC channel is 0 or more actual connections, so yes, if you can reuse clients it would save some resources (unless you're maxing out the bandwidth in which case you'd probably want separate clients)

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: https://arrow.apache.org/docs/java/reference/org/apache/arrow/flight/FlightStream.html#next-- We are getting the stream using `FlightStream.next()`. I couldn't find an option to specify the size of Vector data returned by the stream. So I assume this Vector data size will be determined by the datasource or the flight service.

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: Added details below on how a FlightInfo will be got by authenticating using FlightClient.

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: The current design is such that a FlightClient will be created for a new query and then will be closed after use. Are you proposing to keep this FlightClient open?

lidavidm: It is not strictly necessary but it would be good to keep a pool of clients for frequently used destinations. Of course that depends on the workload. I think this is more an implementation detail though.

elbinpallimalilibm: Added some details on how a connection pool can be maintained of Flight Clients.

RFC-0003-arrow-flight-connector.md
BryanCutler: There isn't a built-in option for this, but it's usually a parameter that could help tune performance on whatever is consuming the Arrow batches. In this case it sounds like it would be useful to tune the conversion from Arrow to a page, depending on the data types and width of the records. So it might be something to keep in mind since it would apply to all datasources.

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Background
+
+Apache Arrow Flight is a new general-purpose client-server framework that simplifies high performance transport of large datasets over network interfaces. https://arrow.apache.org/blog/2019/10/13/introducing-arrow-flight/
+
+Presto currently supports different datasources through connectors for each type of datasource. The connectors can use JDBC or some other protocol for datasource connection. If we use Flight client to connect to the different datasources we can improve the performance of the datasource connection. Also we would only need to develop one Flight client that can be used across different datasources instead of developing individual connectors for each type of datasource.
+
+## Proposed Implementation
+
+A template for Flight connector will be created as `presto-base-arrow-flight` module. Since each implementation of flight server will be different, this base module can be extended to create a flight connector that supports a particular implementation. The template will support only read operations.
```
aditi-pandit, line 25: This section is a bit confusing. What do you mean by template here ?

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,88 @@
...
+    1. The FlightInfo object will be got from FlightClient by authenticating to the FlightClient using a bearer token.
+    2. The concrete implementation of the connector is responsible for generating or providing the bearer token.
+5. Use the FlightInfo object to obtain the list of endpoints we will be retrieving data from. 
+6. Use each endpoint to obtain a Ticket. 
+7. Use the Ticket to obtain a Stream. 
+8. Use the Stream to read data. 
+    1. Get VectorSchemaRoot from Flight stream
+    2. For each column in the result, build a Block from the FieldVector
+    3. Return a page with all the constructed Blocks.
+    4. Return new pages until all the Flight streams are read.
```
aditi-pandit, line 54: For Prestissimo we could directly convert from Arrow to Velox vectors as they are largely compatible. Please can you add some details about that.

aditi-pandit (commented):
Thanks for this RFC. Would be great if you add more detail about Velox/Prestissimo design as well.

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Summary
+
+Create a base Arrow Flight module in Presto which can be extended to create an Arrow Flight connector.
+
+## Background
+
+Apache Arrow Flight is a new general-purpose client-server framework that simplifies high performance transport of large datasets over network interfaces. https://arrow.apache.org/blog/2019/10/13/introducing-arrow-flight/
+
+Presto currently supports different datasources through connectors for each type of datasource. The connectors can use JDBC or some other protocol for datasource connection. If we use Flight client to connect to the different datasources we can improve the performance of the datasource connection. Also we would only need to develop one Flight client that can be used across different datasources instead of developing individual connectors for each type of datasource.
```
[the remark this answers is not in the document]
elbinpallimalilibm: We have individual connectors for each type of datasource like MySQL, Oracle or PostgresQL. If we develop one Flight connector, it can be used across different types of datasources.

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: This module will only provide the base classes needed to create a Flight connector. So this module is a template like `presto-base-jdbc` which can be extended to create a concrete implementation of Flight connector.

elbinpallimalilibm:
> Thanks for this RFC. Would be great if you add more detail about Velox/Prestissimo design as well.

Yes, we are working on the details for Prestissimo. Will add that here.

RFC-0003-arrow-flight-connector.md
aditi-pandit: Sounds good. Thanks !

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,122 @@
...
+
+![Connector splits in Presto](arrow-flight-connector/Connector-splits-in-Presto.png)
+
+#### Class diagram
+
+![Class diagram](arrow-flight-connector/Class-diagram.png)
+
+
+## Prestissimo Implementation
+A single Prestissimo connector will handle all concrete instances of the Java Arrow Flight connector. This is possible because the prestissimo connector does not have to deal with metadata fetch and discovery. Each split wraps a ticket and a list of locations, which is all[^1] that is required to fetch data from the Flight Service in an implementation-agnostic way.
```
aditi-pandit, line 77: This sounds right because we would need only Arrow Flight template level mappings and have a single ArrowFlightDataSource and ArrowFlightSplit.

aditi-pandit, line 77: Its important to consider how the ticket and locations list will be serialized from Presto Java to Prestissimo. https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/presto_protocol/presto_protocol.yml shows the other mappings. Please give some more info on how this will be generated.

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,122 @@
...
+connector.name=arrow-flight
+flight.auth.name=simple-user-pass
+
+# The following are examples of authenticator-specific properties
+# In presto, the concrete module should handle them
+# In prestissimo, the authenticator plugin will handle them
+flight.auth.user=me
+flight.auth.password=itsme
+```
+In the above case, the authenticator registered as `simple-user-pass` will be used to authenticate all Flight requests for that catalog. More authenticators can be created and registered using an interface similar to the interface for connector creation and registration in Velox.
```
aditi-pandit, line 102: Sounds like these authenticators should be registered in PrestoServer class. Is the AuthenticatorClass a pure Velox interface independent of Flight ? Wanted to check if there are flight dependencies for it... If there is then we will need to add dependencies in Velox.

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Conenctor registration
+All concrete implementations of the presto connector will map to the same connector in prestissimo. For example, assume there are two concrete implementaitons of the base module in Java - `arrow-flight-ibm` and `arrow-flight-sql`. In this case, `connector.name` values of `arrow-flight`, `arrow-flight-ibm`, `arrow-flight-sql` would all reference the same connector implementation in prestissimo.
```
aditi-pandit, line 105: Would be great to show how different ArrowFlightDataSource and ArrowFlightSplit objects for different concrete implementations.

aditi-pandit (commented):
Thanks for adding Prestissimo dependencies. Had few questions.

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: @ashkrisk Can you please check the comments on Prestissimo implementation

ashkrisk: The Java implementation looks something like this:
```java
class ArrowSplit {
  byte[] ticket;
  List<String> locations;
}
```
Jackson can serialize `byte[]` and `List<String>`. `byte[]` in particular will be serialized as a base64-encoded string.

The corresponding C++ implementation is quite similar:
```cpp
class FlightSplit {
  std::string ticket;
  std::vector<std::string> locations;
}
```
The only extra detail to take care of is that although the `base64` encoding/decoding is implicit in Java, it has to be handled explicitly in C++.

RFC-0003-arrow-flight-connector.md
ashkrisk: All concrete connectors will use exactly the same (non-templated) classes for `ArrowFlightDataSource` and `ArrowFlightSplit`.

As far as prestissimo is concerned, there is almost no difference between any of these concrete implementations. The only thing that can differ is the `Authenticator` that is in use, but that is stored as a member object.

```cpp
class FlightDataSource {
  // ...
  std::shared_ptr<Authenticator> authenticator;
  // ...
}
```

RFC-0003-arrow-flight-connector.md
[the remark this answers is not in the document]
majetideepak: > Also we would only need to develop one Flight client that can be used across different datasources instead of developing individual connectors for each type of datasource.

This statement is confusing. Isn't Flight client similar to JDBC in terms of capabilities? We have different connectors on top of the JDBC standard because each data source requires customizations.

RFC-0003-arrow-flight-connector.md
ashkrisk: For convenience's sake, the authenticator is a Flight-dependent interface. I'm imagining that it would only be used alongside the arrow flight connector.

The interface itself can certainly be made independent of the flight connector; the tradeoff is that authenticator implementations that need the Flight SDK may end up having to write a few extra lines of boilerplate.

Should we make the interface independent?

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### Proposed design
+
+![Design diagram](arrow-flight-connector/Presto-flight-client.drawio.png)
+
+The Arrow Flight libraries provide a development framework for implementing a service that can send and receive data streams. Flight APIs are used to retrieve the data from the selected table. The steps below need to be followed. 
+
+1. Create a Flight Client using the host, port, ssl information of the Flight service. 
+    1. Maintain a connection pool of Flight clients.
+    2. Reuse a Flight client from the pool that is already connected to the same location as the new connection.
```
majetideepak, line 38: `location` -> `Flight server`
nit: `as the new connection` can be removed.

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,122 @@
...
+    3. Evict connections from the pool using the Least Frequently Used strategy.
+2. Use the Flight Client to connect to the Flight service.
```
majetideepak, line 40: May be add point `1.2`about reuse here.

RFC-0003-arrow-flight-connector.md
majetideepak, line 105: typo `implementaitons` -> `implementations`

RFC-0003-arrow-flight-connector.md
```diff
@@ -0,0 +1,122 @@
...
+
+# The following are examples of authenticator-specific properties
+# In presto, the concrete module should handle them
+# In prestissimo, the authenticator plugin will handle them
+flight.auth.user=me
+flight.auth.password=itsme
+```
+In the above case, the authenticator registered as `simple-user-pass` will be used to authenticate all Flight requests for that catalog. More authenticators can be created and registered using an interface similar to the interface for connector creation and registration in Velox.
+
+### Conenctor registration
```
majetideepak, line 104: typo: `Conenctor` -> `Connector`

RFC-0003-arrow-flight-connector.md
ashkrisk: Added a section with details https://github.com/prestodb/rfcs/blob/2d0740ac0b610057c95026879bd343d5ddc3cedc/RFC-0003-arrow-flight-connector.md?plain=1#L88

RFC-0003-arrow-flight-connector.md
ashkrisk: Added a line to clarify this https://github.com/prestodb/rfcs/blob/2d0740ac0b610057c95026879bd343d5ddc3cedc/RFC-0003-arrow-flight-connector.md?plain=1#L110

RFC-0003-arrow-flight-connector.md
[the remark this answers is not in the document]
elbinpallimalilibm: I meant that we don't have to create different connectors for each type of datasource like we do in JDBC.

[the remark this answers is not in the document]
majetideepak: Why do you think is missing in JDBC that arrow Flight solves?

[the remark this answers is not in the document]
elbinpallimalilibm: Arrow Flight provides better performance than JDBC. The Arrow memory format supports zero-copy reads for lightning-fast data access without serialization overhead.

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: Done

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: Updated

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: @ashkrisk Can you address this?

ashkrisk: I'd already added those back when I originally added the Prestissimo details:
https://github.com/prestodb/rfcs/blob/7fc2ebbfde03716c001972112acf06b6be75a9d3/RFC-0003-arrow-flight-connector.md?plain=1#L81-L86

RFC-0003-arrow-flight-connector.md
elbinpallimalilibm: Mentioned that the flight client can be re-used from a pool.

RFC-0003-arrow-flight-connector.md
aditi-pandit: @ashkrisk : Its better to make the interface independent. And have customizations for flight.

RFC-0003-arrow-flight-connector.md
tdcmeehan: Yes--the implementation must not presume that the underlying flight service accepts SQL.  Please remove any reference to queries and SQL>

RFC-0003-arrow-flight-connector.md
ashkrisk: I'm not sure if there is much value in making the interface independent if it's not getting used elsewhere. Perhaps we can resolve this during code review once you've had a chance to see what the current implementation looks like and if you feel it's the sort of thing that you might want to use outside the Flight connector?

aditi-pandit: @ashkrisk : My main concern with adding Flight related code in the generic interface is because flight libraries would compile with the velox core. We try to avoid dependent libraries as much as possible and limit them to only when the flight connector is needed.

ashkrisk: @aditi-pandit the Authenticator interface is defined within the folder for the Arrow Flight connector and is only enabled/compiled when Prestissimo is built with `PRESTO_ENABLE_ARROW_FLIGHT_CONNECTOR=ON`. Authenticator implementations would have a dependency on the Arrow Flight libraries, but if you don't build the connector (and hence don't build any authenticators for it) you don't need the Flight libraries at all.

Does this resolve your concern?

aditi-pandit: @ashkrisk : This sounds a bit odd. Though we can look at your code to resolve this.

tdcmeehan (approved):
LGTM.  Any other feedback @BryanCutler @aditi-pandit?

tdcmeehan (commented):
Oh, but please update this to RFC-0004, there's already an RFC-0003

RFC-0004-arrow-flight-connector.md
```diff
@@ -0,0 +1,130 @@
+# **RFC0003 for Presto**
```
tdcmeehan, line 1:
```suggestion
# **RFC-0004 for Presto**
```

tdcmeehan (changes requested):

tdcmeehan (approved):

PR #12, opened 2024-06-11
Add RFC for supporting distributed procedure

hantangwangd:
Propose design to expand the current procedure architecture in presto, support defining, registering and calling procedures which need to be executed in a distributed way.

Besides, in order to demonstrate as an example and to figure out the functional boundaries among the different architecture levels, also describe the design for Iceberg to support distributed procedure as well as the design of a specific distributed procedure `rewrite_data_files`.


rschlussel (commented):
This is an interesting feature. do you have thoughts on integrating this for cpp workers as well?

hantangwangd:
@rschlussel Thanks for the comment. Yes, of course we should integrate this for cpp workers if it's proved to be useful and needed.

I think we can take a two-step approach for this. Firstly, we support it on java workers, confirm the feasible of the entire architecture and figure out the best functional boundary division. Then, after that, we can support it on cpp workers following the design and implementation path on java workers. What's your opinion?

RFC-0004-support-distributed-procedure.md
```diff
@@ -0,0 +1,397 @@
...
+
+
+5. Similar to statements such as `create table as`/`insert`/`delete`/`refresh material view` that involve distributed processing, two related SPI methods are defined for `call distributed procedure` statement in the metadata and connector metadata interfaces. `beginCallDistributedProcedure` is used for preparation work before the start of distributed scheduling. And `finishCallDistributedProcedure` is used for transaction commitment after the completion of distributed writing.
+
+
+6. As for a specific connector (such as Iceberg), in the implementation logic of method `beginCallDistributedProcedure` and `finishCallDistributedProcedure` in `ConnectorMetadata`, in addition to accomplish the common logic for this connector (such as starting a transaction, building a transaction context, committing a transaction, etc.), it should also resolve the specified distributed procedure and call its relevant method to execute the procedure-customized logic.
```
aditi-pandit, line 119: Native vs Java runtimes begin to diverge significantly at this point. For native runtimes its best if the distributed procedure call method is in C++. But that would mean older java procedures would need a rewrite.  Do you have any particular ideas on how you will handle this for Native engine ?

RFC-0004-support-distributed-procedure.md
```diff
@@ -0,0 +1,397 @@
...
+```
+
+At present, it can be regarded as a simplified version of `TableWriterNode`, and holds `CallDistributedProcedureTarget`, a subclass of `WriterTarget` used to maintain information about the procedure and the call parameters:
+
+```java
+public static class CallDistributedProcedureTarget
+            extends WriterTarget
+{
+    private final QualifiedObjectName procedureName;
+    private final Object[] procedureArguments;
```
aditi-pandit, line 205: Presume the procedureArguments have types. There could be a difference in the types supported by native vs java execution. Would be great to clarify about that.

RFC-0004-support-distributed-procedure.md
```diff
@@ -0,0 +1,397 @@
...
+
+During the analyzing and logical planning phase, the `call distributed procedure` statement will be planned into a logical plan with the following shape:
+
+```text
+TableScanNode -> FilterNode -> CallDistributedProcedureNode -> TableFinishNode -> OutputNode
+```
+
+Among them, `TableScanNode -> FilterNode` defines the data to be processed. It's based on the target table determined by `schema` and `table_name` in the parameters, as well as the possibly existing filter conditions.
+
+The `CallDistributedProcedureNode -> TableFinishNode` structure is similar to the `TableWriterNode -> TableFinishNode` structure, used to perform distributed data manipulation and final unified submission behavior.
```
aditi-pandit, line 221: Would be great to give more details on what row/control messages are exchanged between the CallDistricutedProcedureNode and TableFinishNode. e.g. TableWriteNode provides protocol of PartitionUpdates and CommitContext that is shared between the two.

RFC-0004-support-distributed-procedure.md
```diff
@@ -0,0 +1,397 @@
...
+
+As for a specific connector (such as Iceberg), in the implementation logic of method `beginCallDistributedProcedure` and `finishCallDistributedProcedure` in `ConnectorMetadata`, in addition to accomplish the common logic for this connector (such as starting a transaction, building a transaction context, committing a transaction, etc.), it should also resolve the specified distributed procedure and call its relevant method to execute the procedure-customized logic.
+
+#### 6. Iceberg connector's support for distributed procedure
+
+In Iceberg, we often need to record the original data files that have been scanned and rewritten during the execution of table data operations (including deleted files that have been fully applied), and in the final submission, combine the newly generated data files due to rewriting to make some changes and transaction submissions at the metadata level.
```
aditi-pandit, line 314: Iceberg in Native engine uses HiveConnector itself. The HiveConnector in Native engine handles both Hive and Iceberg splits. Does HiveConnector support distributed procedure ? How will it be reused or enhanced for Iceberg ?

aditi-pandit (commented):
Thanks @hantangwangd for this RFC. Had couple of comments mainly related to Native execution related impact.

RFC-0004-support-distributed-procedure.md
```diff
@@ -0,0 +1,397 @@
...
+   * Acquire and use procedure registry in `presto-analyzer` and `connectors` module
+
+
+2. Define a new query type `CALL_DISTRIBUTED_PROCEDURE`, and associate the call distributed procedure statement with this type during preparer phase. So that this kind of statements would fall into the path of `SqlQueryExecutionFactory.createQueryExecution(...)`. The generated `SqlQueryExecution` object can utilize the existing distributed querying, writing, and final committing mechanism to achieve distributed execution for procedures.
+
+
+3. Add a new plan node type: `CallDistributedProcedureNode`. During the analyzing and logical planning phase, construct a logical plan with the following shape for `call distributed procedure` statement:
+
+    ```text
+    TableScanNode -> FilterNode -> CallDistributedProcedureNode -> TableFinishNode -> OutputNode
```
jaystarshot, line 104: Not sure I understand correctly, but if we are going to support a general distributed procedure then why does it have to be between tableScan and TableFinish operators?
If its not for a general case but specific to table layouts, then why not just a custom operator implementation?

RFC-0004-support-distributed-procedure.md
```diff
@@ -0,0 +1,397 @@
...
+
+* Dong Wang
+
+## Related Issues
+
+* https://github.com/prestodb/presto/issues/20425
+
+## Summary
+
+Propose design to expand the current procedure architecture in presto, support defining, registering and calling procedures which need to be executed in a distributed way.
```
jaystarshot, line 14: Can you also define what you mean by procedure?

RFC-0004-support-distributed-procedure.md
hantangwangd: Thanks for the comment. Actually, `beginCallDistributedProcedure` and `finishCallDistributedProcedure` in `ConnectorMetadata` are all invoked in coordinator, so native worker do not need to handle them, that is, there is no need in native worker end to resolve the distributed procedures and invoke them.

The non-coordinator worker's responsibility in `call distributed procedure` is similar to that of `insert into` or `create table as`. That is, local planning `CallDistributedProcedureNode` to a `TableWriterOperator` with `ExecutionWriterTarget`, getting corresponding `ConnectorPageSink` based on this writer target, executing the data writing and finally returning the fragments page to table finish stage.

So I think there is no need to consider this issue, what do you think? Any misunderstand please let me know.

RFC-0004-support-distributed-procedure.md
hantangwangd: As I understand, `CallDistributedProcedureTarget` is a subclass of `WriterTarget` which is used only in coordinator. That is, it's always used in java execution environment.

`ExecuteProcedureHandle` which is a subclass of `ExecutionWriterTarget` will be sent to worker. The handling of it is similar to `InsertHandle` or `CreateHandle`.

RFC-0004-support-distributed-procedure.md
hantangwangd: Sure, it's a good suggestion. I will supplement this part of the content.

RFC-0004-support-distributed-procedure.md
hantangwangd: As `beginCallDistributedProcedure` and `finishCallDistributedProcedure` in `ConnectorMetadata` are all invoked in coordinator, I think the main change for hive to support distributed procedures is in java implementation.

Hive should implement `beginCallDistributedProcedure` and `finishCallDistributedProcedure` in `ConnectorMetadata`, customize it's own preparation and submission logic, and then implement and register its own distributed procedures. All these jobs could be done in java only.

So I think the main change in worker end logic which need to support native c++ is that it should generate and provide `ConnectorPageSink` based on the newly added `ConnectorDistributedProcedureHandle`. Please let me know if there are any omissions.

RFC-0004-support-distributed-procedure.md
hantangwangd: Thanks for the comment. As described in this RFC, compared to existing coordinator-only procedures, these newly expanded kind of procedures which need to be executed distributively involve operations on table's data other than metadata. For example, these procedures can rewrite table data, merge small data files, sort table data, repartition table data etc. So the common action process of them includes reading data from the target table, doing some filtering and translation, writing the translated data to files and submitting the changes in metadata level. So I think the general distributed procedure could be planned to a plan tree like this. (Or adding some other plan nodes like JoinNode/AggregationNode etc. for functional expansion in future)

RFC-0004-support-distributed-procedure.md
hantangwangd: Thanks for the suggestion. Strictly speaking, maybe they should be called connector specific system procedures. I'll add some explanations here.

RFC-0004-support-distributed-procedure.md
hantangwangd: > why not just a custom operator implementation?

After optimization and plan fragmentation, we should be able to shuffle data between stages to utilize the capabilities of the entire cluster.

hantangwangd:
This RFC is ready for review, the comments are all handled, and the relevant implementation can be viewed here: https://github.com/prestodb/presto/pull/22659. Please take a look when convenient, thanks! cc: @rschlussel @aditi-pandit @jaystarshot @tdcmeehan @ZacBlanco @yingsu00

RFC-0006-support-distributed-procedure.md
```diff
@@ -0,0 +1,418 @@
...
+## Proposed Implementation
+
+### Overview
+
+1. Distributed procedure is used for processing of actual data besides metadata for a target table, such as data rewriting, merging, sorting or repartitioning. We need to re-factor and expand `Procedure/ProcedureRegistry` architecture, so that we can:
+   * Define a distributed procedure and register it into procedure registry
+   * Acquire and use procedure registry in `presto-analyzer` and `connectors` module
+
+
+2. Define a new query type `CALL_DISTRIBUTED_PROCEDURE`, and associate the call distributed procedure statement with this type during preparer phase. So that this kind of statements would fall into the path of `SqlQueryExecutionFactory.createQueryExecution(...)`. The generated `SqlQueryExecution` object can utilize the existing distributed querying, writing, and final committing mechanism to achieve distributed execution for procedures.
```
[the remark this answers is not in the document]
hantangwangd: Thanks for your suggestion. I will supplement this part of the content. The following is a brief explanation.

We propose the syntax of `Call statement` for the reason described in line 24 to line 28 (https://github.com/prestodb/rfcs/pull/12/files#diff-7655a74167c417d09fba4cc48ec39232f65d7056a449e1cb09d51664f4fe1b7eR24-R28), for example:

```
CALL iceberg.system.rewrite_data_files('db', 'sample');
CALL iceberg.system.rewrite_data_files(schema => 'db', table_name => 'sample');
CALL iceberg.system.rewrite_data_files('db', 'sample', 'partition_key = 1');
CALL iceberg.system.rewrite_data_files(schema => 'db', table_name => 'sample', filter => 'partition_key = 1');
```

For all distributed procedures, the implementation of their base class defaults to including parameters `schema`, `table_name` and `filter`. Among them, `schema` and `table_name` are required parameters, which specify the target table to be processed by this distributed procedure. While `filter` is an optional parameter, which indicates the filtering conditions for the data to be processed.

The specific procedure implementations can expand the parameter list, but they must include the required parameters `schema` and `table_name`.

RFC-0006-support-distributed-procedure.md
```diff
@@ -0,0 +1,418 @@
...
+
+
+3. Add a new plan node type: `CallDistributedProcedureNode`. During the analyzing and logical planning phase, construct a logical plan with the following shape for `call distributed procedure` statement:
+
+    ```text
+    TableScanNode -> FilterNode -> CallDistributedProcedureNode -> TableFinishNode -> OutputNode
```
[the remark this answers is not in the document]
hantangwangd: It's a regular table representing the target table. As mentioned above, the target table is specified through the parameters `schema` and `table_name`.

RFC-0006-support-distributed-procedure.md
```diff
@@ -0,0 +1,418 @@
...
+    ```
+	
+    Wherein, `CallDistributedProcedureNode` holds `CallDistributedProcedureTarget`, which is a subclass of `WriterTarget`, used to maintain the procedure information and related call parameters information.
+
+
+![Distributed_procedure_architecture](RFC-0006/distributed_procedure.png)
+
+
+4. The optimizing, segmenting, group execution tagging, and local planning of `CallDistributedProcedureNode` are similar to `TableWriterNode`. And it would be ultimately local planned to a `TableWriterOperator` (which holds a specific type of `ExecutionWriterTarget` subclass related to `call distributed procedure` statement). When creating a `PageSink` to execute data writing, a corresponding `ConnectorPageSink` will be generated based on the specific subclass and property values of `ExecutionWriterTarget` that are actually held by `TableWriterOperator`.
```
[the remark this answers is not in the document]
hantangwangd: I'm not so familiar with Prestissimo implementation. Are we currently reusing the implementation of `ConnectorPageSink` from Hive in native worker? If so, yes maybe it's better to implement an Iceberg own `ConnectorPageSink`.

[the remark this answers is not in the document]
hantangwangd: The detailed description for `ExecuteProcedureHandle` (which is an implementation class of interface `ExecutionWriterTarget`) and it's members is in line 252 to line 273 (https://github.com/prestodb/rfcs/pull/12/files#diff-7655a74167c417d09fba4cc48ec39232f65d7056a449e1cb09d51664f4fe1b7eR252-R273). The definitions of its fields and members are similar to `InsertHandle`.

Currently there is no specific logic that needs to be implemented through code in `ExecuteProcedureHandle`, so in order to support `ExecuteProcedureHandle` in native, we can basically refer to the implementation of `InsertHandle`. The only difference is that an additional field will appear in `ExecuteProcedureHandle`: `QualifiedObjectName procedureName`.

RFC-0006-support-distributed-procedure.md
```diff
@@ -0,0 +1,418 @@
...
+
+
+5. Similar to statements such as `create table as`/`insert`/`delete`/`refresh material view` that involve distributed processing, two related SPI methods are defined for `call distributed procedure` statement in the metadata and connector metadata interfaces. `beginCallDistributedProcedure` is used for preparation work before the start of distributed scheduling. And `finishCallDistributedProcedure` is used for transaction commitment after the completion of distributed writing.
```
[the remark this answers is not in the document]
hantangwangd: OK, will add this.

RFC-0006-support-distributed-procedure.md
```diff
@@ -0,0 +1,418 @@
...
+}
+```
+
+At present, it can be regarded as a simplified version of `TableWriterNode`, and holds `CallDistributedProcedureTarget`, a subclass of `WriterTarget` used to maintain information about the procedure and the call parameters:
+
+```java
+public static class CallDistributedProcedureTarget
+            extends WriterTarget
+{
+    private final QualifiedObjectName procedureName;
```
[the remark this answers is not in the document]
hantangwangd: Firstly, in the current design, distributed procedures serve as Connector specific built-in system procedures, therefore they do not need to be dynamically loaded, but only need to be statically pre-defined and loaded. Dynamic loading is more like what need to be done when supporting `UDF` or `UDP`, which is beyond the scope of this proposal. And we can consider how to support it if necessary in the future.

Secondly, as discussed above, the methods of a `DistributedProcedure` will always be called in Java environment. Even when executing the final submission, the method `FinishCallDistributedProcedure.finish(...)` will be executed in operator `TableFinishOperator` in the Java worker inside `coordinator` as well. So it seems that there is currently no need to consider dynamic loading on C++ environment. Please let me know if there are any omissions or misunderstandings.

RFC-0006-support-distributed-procedure.md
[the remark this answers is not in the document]
hantangwangd: Done, added in https://github.com/prestodb/rfcs/pull/12/files#diff-7655a74167c417d09fba4cc48ec39232f65d7056a449e1cb09d51664f4fe1b7eR131-R132.

RFC-0006-support-distributed-procedure.md
```diff
@@ -0,0 +1,431 @@
...
+> Besides, it can also contain an optional parameter `filter` with the default value of `TRUE`, which indicates the filtering conditions for the data to be processed.
+> A specific procedure implementation can define the parameter list on its own, but it must include the required parameters `schema` and `table_name`, and can optionally use parameter `filter` to narrow down the range of data to be processed.
+
+2. Define a new query type `CALL_DISTRIBUTED_PROCEDURE`, and associate the call distributed procedure statement with this type during preparer phase. So that this kind of statements would fall into the path of `SqlQueryExecutionFactory.createQueryExecution(...)`. The generated `SqlQueryExecution` object can utilize the existing distributed querying, writing, and final committing mechanism to achieve distributed execution for procedures.
+
+
+3. Add a new plan node type: `CallDistributedProcedureNode`. During the analyzing and logical planning phase, construct a logical plan with the following shape for `call distributed procedure` statement:
+
+    ```text
+    TableScanNode -> FilterNode -> CallDistributedProcedureNode -> TableFinishNode -> OutputNode
```
yingsu00, line 117: The operation may not necessarily update the real data files. For example, We may want to run `CALL system.clean_data_cache('ALL_WORKERS');` to just call `AsyncDataCache::shutdown() ` on all Presto C++ workers. What would the plan look like?

yingsu00 (commented):
@hantangwangd Is there a working prototype or a PR now?

RFC-0006-support-distributed-procedure.md
yingsu00: cc @agrawalreetika

hantangwangd:
> Is there a working prototype or a PR now?

Yes, there is already a PR, [#22659](https://github.com/prestodb/presto/pull/22659). Will rebase it to the newest commit list as soon as possible.

tdcmeehan (commented):
@hantangwangd thanks for your perseverance on this RFC.

It would be nice to sketch out the work required to execute distributed procedures in C++ clusters.  It would seem like it's the following:

1) Add `CallDistributedProcedureNode` to the protocol
2) Add a new method `PrestoToVeloxConnector.toVeloxInsertTableHandle` that interprets the handle and creates a Velox insert handle

tdcmeehan (commented):
LGTM.  Can you update the RFC number in the file name and also add it to the title, similar to other RFCs?

hantangwangd:
@tdcmeehan Sure, I have updated the RFC number of the file name to RFC-0021 and added it to the title. Please take a look when you have a chance. Thanks a lot!

RFC-0021-support-distributed-procedure.md
```diff
@@ -0,0 +1,490 @@
...
+
+> Both `beginCallDistributedProcedure` and `finishCallDistributedProcedure` are all invoked in coordinator. More precisely, `finishCallDistributedProcedure` will always be invoked by operator `TableFinishOperator` in the Java worker inside coordinator.
+> That is to say, they will always be called in Java environment.
+
+6. As for a specific connector (such as Iceberg), in the implementation logic of method `beginCallDistributedProcedure` and `finishCallDistributedProcedure` in `ConnectorMetadata`, in addition to accomplish the common logic for this connector (such as starting a transaction, committing a transaction, etc.), it should also construct the procedure context, resolve the specified distributed procedure and call its relevant method to execute the procedure-customized logic.
+
+### Detailed description
+
+#### 1. Re-factor Procedure/ProcedureRegistry
+
```
aditi-pandit, line 145: Don't think its clear about how we discover these procedures on server startup... Are these like plugins ?

hantangwangd: @aditi-pandit thanks for your feedback. In summary, it extends and reuses the existing procedure registration and discovery mechanism, so it follows the same pattern as existing Procedures. See the registration of the newly added `RewriteDataFilesProcedure` [here](https://github.com/hantangwangd/presto/blob/support_call_distributed_procedure_part2/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergCommonModule.java#L177-L189).

tdcmeehan (approved):

PR #13, opened 2024-06-12
Add expression evaluation RFC

tdcmeehan:


RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Summary
+
+Constant folding is a technique used in compilers to evaluate constant expressions at compile time. This RFC proposes to add constant folding to Presto's expression evaluation engine. This will allow Presto to evaluate constant expressions at query planning time, reducing the amount of work that needs to be done at query execution time.
+
+RFC-0003 proposed a mechanism to add constant folding support via remote function evaluation.  This RFC has a similar goal, but represents an improvement over that approach.
+
+## Background
+
+Unfenced functions are functions which execute within the same process as the underlying evaluation engine.  Typically, this term is used in the context of user-defined functions, and it is mean to contrast with fenced functions, which execute in a separate process.  Unfenced functions are more efficient than fenced functions, but they are also more dangerous, as they can crash the process in which they are running.
```
aditi-pandit, line 21: Nit : "meant" instead of mean

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+
+Currently in Presto C++, there is no constant folding support for functions which are written in C++.  In practice, constant folding works for functions which have a corresponding Java implementation, but not for functions which are implemented in C++.  It works by essentially just using a forked set of functions written in Java to evaluate the constant expressions, and then during execution relying on the C++ functions to evaluate the non-constant expressions.  This has a few drawbacks:
+
+1) It requires a forked set of functions to be written in Java, which is a maintenance burden.
+2) An unfenced function written in C++ must have a corresponding Java implementation in order to work properly.
+
+RFC-0003 proposed a mechanism to both add a function registry to allow C++ implemented functions to be exposed to the coordinator without a corresponding Java implementation, and to add constant folding support via remote function evaluation.  This RFC has a similar goal, but represents an improvement over that approach.
+
+In this RFC, we propose to optimize row expressions as a whole, not just function calls.  This has a few advantages:
```
aditi-pandit, line 34: Would be great to illustrate the functionality with an example.

aditi-pandit (commented):
@tdcmeehan : Had a quick look. Looks good overall. Had just one comment.

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+
+A key feature of unfenced functions is that they can be used to implement constant folding.  This is because the function can be evaluated at query planning time, and the result can be used as a constant in the query plan.  This can be a significant performance improvement, as it can reduce the amount of work that needs to be done at query execution time.  For example, if an unfenced function is evaluated over a partition column, the result can be used to prune partitions before the query is executed.
+
+Currently in Presto C++, there is no constant folding support for functions which are written in C++.  In practice, constant folding works for functions which have a corresponding Java implementation, but not for functions which are implemented in C++.  It works by essentially just using a forked set of functions written in Java to evaluate the constant expressions, and then during execution relying on the C++ functions to evaluate the non-constant expressions.  This has a few drawbacks:
```
Yuhta, line 27: C++ do have constant folding during operator initialization.  The expression is evaluated once per node per query.

tdcmeehan: To clarify, while C++ does have constant folding support during eval, what is needed is C++ constant folding support during runtime (at the very least to simplify expressions over partition columns, but also because the coordinator rightly will attempt to simplify all expressions during optimization).  Let me clarify this in the doc.

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+## [Optional] Metrics
+
+This is a 0 to 1 feature, and its success will be largely dictated by two metrics:
+
+1) The performance of evaluating expressions, particularly large, complicated expressions.  This can be measured by comparing the time it takes to evaluate a query with and without the new SPI.
+2) The ability to perform constant folding in C++ clusters with custom functions which are only registered in C++.
+
+## [Optional] Other Approaches Considered
+
+A per-function approach was initially considered, however this is chosen in favor of that because not only is the API cleaner, we also expect that it can scale better.
```
pedroerp, line 71: For context, we created something similar in Velox, but it's restricted to one function call since pushing down entire expressions bring a ton of complexity regarding null handling, encodings, special forms, etc. I think if you're only planning to push this down to Presto C++ it should be ok, since it has a full implementation of all that stuff by definition.

There is support in Velox for people to register "remote" functions, and a thrift API for them to implement it - though the API could easily be made plugable:

https://github.com/facebookincubator/velox/blob/main/velox/functions/remote/if/RemoteFunction.thrift

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+### [Optional] Goals
+
+* Allow for a pluggable implementation of expression evaluation and use this in the Presto optimizer for constant folding
+* Allow connector authors to use this plugin to optimize their own expressions (for example, for filter pushdown)
+
+### [Optional] Non-goals
+
+## Proposed Implementation
+
+A new SPI will be added to Presto to allow for a pluggable implementation of expression evaluation.  A default implementation of this will be added which delegates to the existing expresion evaluation code, but this may be overridden by a custom plugin.
```
pedroerp, line 48: @mbasmanova @kgpai @kagamiori I wonder if somehow this SPI could also be used to validate fuzzed expressions in Velox using the Presto Java eval. We could probably use the same API but in a different direction - having the C++ code calling a Java backend to evaluate the expression. 

@tdcmeehan could this be also supported in by this API/SPI?

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+
+* Tim Meehan
+
+## [Related Issues]
+
+* RFC-0003
+
+## Summary
+
+Constant folding is a technique used in compilers to evaluate constant expressions at compile time. This RFC proposes to add constant folding to Presto's expression evaluation engine. This will allow Presto to evaluate constant expressions at query planning time, reducing the amount of work that needs to be done at query execution time.
```
mbasmanova, line 15: > This RFC proposes to add constant folding to Presto's expression evaluation engine.

Doesn't Presto already have constant folding?

RFC-0004-expression-eval.md
mbasmanova, line 27: Another issue is that Java and C++ function have slightly difference semantics (regex, json, upper, lower, date_format, cast(double as varchar), etc.). If a function is evaluated in Java during constant folding and then again in C++ the user may experience a "mix" of semantics within a single query.

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+
+Currently in Presto C++, there is no constant folding support for functions which are written in C++.  In practice, constant folding works for functions which have a corresponding Java implementation, but not for functions which are implemented in C++.  It works by essentially just using a forked set of functions written in Java to evaluate the constant expressions, and then during execution relying on the C++ functions to evaluate the non-constant expressions.  This has a few drawbacks:
+
+1) It requires a forked set of functions to be written in Java, which is a maintenance burden.
+2) An unfenced function written in C++ must have a corresponding Java implementation in order to work properly.
+
+RFC-0003 proposed a mechanism to both add a function registry to allow C++ implemented functions to be exposed to the coordinator without a corresponding Java implementation, and to add constant folding support via remote function evaluation.  This RFC has a similar goal, but represents an improvement over that approach.
```
mbasmanova, line 32: Where can I find RFC-0003?

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+
+1) It allows for constant folding to be applied to expressions which are not function calls, such as special form expressions.
+2) It can reduce the overhead of constant folding by allowing the evaluation engine to evaluate the expression as a whole, rather than evaluating each function call individually.
+
+### [Optional] Goals
+
+* Allow for a pluggable implementation of expression evaluation and use this in the Presto optimizer for constant folding
+* Allow connector authors to use this plugin to optimize their own expressions (for example, for filter pushdown)
+
+### [Optional] Non-goals
```
mbasmanova, line 44: Should we mention that making this as performant and current approach is a non-goal? Or should we have some goals on "max" execution latency introduced by this proposal?

RFC-0004-expression-eval.md
mbasmanova: @pedroerp Pedro, it is not clear to me what would be the benefit of this over the current approach or using PrestoQueryRunner. Would you clarify?

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+
+A new plugin will be implemented which will optimize row expressions by serializing them and sending them to the Presto sidecar for evaluation.  The Presto sidecar will have a new endpoint which can take in a serialized row expression and return the result of evaluating it.
+
+> Endpoint: /v1/evaluate
+> 
+> HTTP verb: POST
+> 
+>Body: JSON list of serialized row expressions
+>
+> Response: JSON list of optimized serialized row expressions
```
mbasmanova, line 58: I expected the response to be a list of constant values that represent the results of evaluating the expressions. Is this not the case? Why are you saying "optimized ... expressions" and not "values"?

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+ 
+The new plugin will be integrated into the Presto optimizer through a new optimizer rule similar to `RowExpressionOptimizer`, except which uses the new SPI to evaluate the expressions.  There will be a feature toggle which can enable or disable the use of this new SPI, however eventually it will become mandatory to use it.
+
+## [Optional] Metrics
+
+This is a 0 to 1 feature, and its success will be largely dictated by two metrics:
+
+1) The performance of evaluating expressions, particularly large, complicated expressions.  This can be measured by comparing the time it takes to evaluate a query with and without the new SPI.
```
mbasmanova, line 66: Actually, it might be more important to measure overhead of constant-folding simple expressions like 1 + 2.

RFC-0004-expression-eval.md
mbasmanova, line 71: Love it. Thanks!

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+2) The ability to perform constant folding in C++ clusters with custom functions which are only registered in C++.
+
+## [Optional] Other Approaches Considered
+
+A per-function approach was initially considered, however this is chosen in favor of that because not only is the API cleaner, we also expect that it can scale better.
+
+## Adoption Plan
+
+- This feature will be disabled by default.
```
mbasmanova, line 75: Can we enable it by default when native_execution is true?

RFC-0004-expression-eval.md
```diff
@@ -0,0 +1,80 @@
...
+- After a period of time, the default will be to enable it.
+
+## Test Plan
+
+In addition to using the existing test coverage for expression evaluation to catch bugs in the expression optimizer, we will also use a fuzzer to generate random expressions and evaluate them with and without the new SPI to ensure that the results are the same.
```
mbasmanova, line 80: It would be nice to add a section about performance testing and also add e2e tests for the optimizer with this new functionality enabled. Specifically, it would be nice to make sure errors are handled properly, e.g. sidecar crashing, not responding, taking very long time to respond, not able to evaluate a function, etc.

mbasmanova (commented):
@tdcmeehan Tim, thank you for working on this. Some comments.

RFC-0004-expression-eval.md
rschlussel: https://github.com/prestodb/rfcs/blob/main/RFC-0003-native-spi.md

RFC-0004-expression-eval.md
tdcmeehan: I think it could, but we are actually looking to see if we can use the Velox expression fuzzer to do an A/B test between the Java-only expression optimizer and the Presto C++-backed expression optimizer, since a lot of the work has already been done with this fuzzer and we would prefer not to write it from scratch.

RFC-0004-expression-eval.md
tdcmeehan: Yes, I meant something along the lines of the following: "This RFC proposes to add pluggable constant folding support to Presto's optimizer so that the underlying evaluation engine can simplify expressions."  I'll update the doc, and concisely mention the motivation here as well.

RFC-0004-expression-eval.md
tdcmeehan: This is a good point which is another motivation.  I will add it.

RFC-0004-expression-eval.md
tdcmeehan: For most cases, they will be values, however I think optimized expression better conveys that there are cases when they won't be simple constants (as I somehow understand from the word value).  Some examples I have in mind:

* CAST(int_column AS INT) => int_column
  * We would expect the expression optimizer to remove the unneeded cast
* IF(condition, 123, 0/0) => CAST(FAIL(...) AS INTEGER) -- also, similar for CASE statement
  * We would expect the expression optimizer to only fail the query if the false condition is reached
* COALESCE(val1, val2, val1, val3) => COALESCE(val1, val2, val3)
  * We would expect the expression optimizer to remove the unneeded and redundant statements in the coalesce

I planned to add the mechanism of how we determine what may be constant folded, and how we batch them and send them to the sidecar, as a part of this document (I am returning from PTO and wanted to get this out in draft state before I left).

RFC-0004-expression-eval.md
tdcmeehan: Agreed, I will add a metric that also mentions the overhead for simple expressions, as we'd want to prove it's negligible.

RFC-0004-expression-eval.md
mbasmanova: @tdcmeehan Tim, thank you for clarifying. I feel this is going beyond constant folding and it not supported in Velox / Prestissimo. I wonder if it makes sense to reduce the scope of this proposal to just constant folding, e.g. taking an expression with no column references and returning a value.

tdcmeehan: While we wanted to push down as much as possible to the underlying evaluation engine, since we don't make any distinction between general expression optimization and the more specific action of reducing constant expressions to literal values, I understand that this might be considered to be out of scope for Velox, and we can make this distinction.

RFC-0004-expression-eval.md
mbasmanova: @pedroerp My understanding is that PQR already works for any query, but it needs to be integrated into Expression Fuzzer in a similar way it was integrated into Aggregation Fuzzer.

RFC-0004-expression-eval.md
tdcmeehan: @mbasmanova Consider the following expressions with references to other columns:

* `ABS(COALESCE(123, column1))` => `123`
* `IF(true, 123, column2)` => `123`

Do we consider the simplification of these expressions within scope for constant folding in Velox?

tdcmeehan: I'm wondering if it's possible to avoid having to make multiple passes for each expression in the case of nested expressions which are able to constant fold into values.

If we take the above examples and combine them in a way:

`CppFlipSign(IF(MyCppFunc(COALESCE(123, column1)) < 456, 456, column2))`

Then whether or not Velox supports constant folding of special form expressions where there is a reference to a different column seems to make a big difference.  If Velox is able to constant fold those examples, then we can send the whole expression in one call.  If it can't, then it would require multiple passes between Presto and Velox to constant fold the expression.

Presto: `COALESCE(123, column1)` => `123` (contains a column reference, yet because of the properties of this special form expression, we can infer that we can just collapse it to 123)
Velox: `MyCppFunc(123)` => `123` (suppose for simplicity this just returns the input)
Presto: `IF(123 < 456, 456, column2)` => `456` (contains a column reference, yet because of the properties of this special form expression, we can infer that we can just collapse it to 456)
Velox: `CppFlipSign(456)` => `-456`

In this case, at least two calls are unavoidable, because Velox cannot determine that the special form expressions are constant foldable, and Presto can't execute the function by design.  I briefly looked at the code in Velox, can confirmed that it appears we only check if something is constant foldable if all of the inputs are constant expressions.  However, I'm wondering if we should add more sophistication to this process?  I am guessing that the original motivation behind constant folding expressions with only constant inputs would also apply to inferring the unique properties of special form expressions and constant folding such expressions where possible?

mbasmanova: @tdcmeehan Tim, these are all good questions / points. My initial thinking is to restrict Velox to "true" constant folding, e.g. evaluating expressions that do not reference any columns and producing a "value" as a result. I see how this may result in extra trips between coordinator and side-car, but I feel the alternative adds a lot of complexity. If you feel it is important to push this logic into the side-car, than, perhaps, a next step would be to describe all the optimizations we would expect to implement and then consider whether it makes sense to do that in Velox (or Prestissimo layer) or not. 

What do you think?

tdcmeehan: I think this design works best when there's only a single hop, so it seems like the question is do we do this in Prestissimo or Velox.  I'll mention the optimizations needed to achieve general constant folding support which is at parity with Java (which includes expressions that contain column references), and leave it as a detail to be determined where the code lives.

mbasmanova: Sounds good. We may need to build some optimizer logic in C++ in Prestissimo to make this work. Note that conversion of expressions to Velox is one-way. It is not possible to go back cleanly.

Yuhta: I don't think multiple passes in this use case is really an issue performance-wise.  Making the constant-folding in Velox more powerful might be useful in other use cases on execution side though.

RFC-0004-expression-eval.md
tdcmeehan: This is in the context of unfenced or intra process functions where we can process them early in the query optimizer.  Processing remote functions in the optimizer might introduce unacceptable latency, so I think we'll defer that to execution.

tdcmeehan:
@Yuhta @mbasmanova @pedroerp @aditi-pandit thank you very much for your thoughtful feedback.  I have marked this RFC as ready for review and I have addressed your many points.  Please let me know if you have more suggestions or feedback, thank you!

RFC-0004-expression-eval.md
tdcmeehan: I have an example now in the proposed implementation section.

aditi-pandit (approved):

RFC-0005-expression-eval.md
```diff
@@ -0,0 +1,217 @@
...
+
+A key feature of unfenced functions is that they can be used to implement constant folding.  This is because the function can be evaluated at query planning time, and the result can be used as a constant in the query plan.  This can be a significant performance improvement, as it can reduce the amount of work that needs to be done at query execution time.  For example, if an unfenced function is evaluated over a partition column, the result can be used to prune partitions before the query is executed.
+
+Currently in Presto C++, there is no constant folding support during query optimization for functions which are written in C++.  In practice, constant folding works for functions which have a corresponding Java implementation, but not for functions which are implemented in C++.  It works by essentially just using a forked set of functions written in Java to evaluate the constant expressions, and then during execution relying on the C++ functions to evaluate the non-constant expressions.  This has a few drawbacks:
```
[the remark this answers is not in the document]
tdcmeehan: I have reworded it, please let me know what you think.

aditi-pandit (approved):
Thanks for the revisions !

PR #14, opened 2024-07-01
RFC stats propagation for functions/UDFs

ScrapCodes:
cc @aaneja @ZacBlanco @aditi-pandit

### DIFF
--- a/RFC-0004-functions-stats.md
+++ b/RFC-0004-functions-stats.md
@@ -1,11 +1,11 @@
 # **RFC0004 for Presto**
 
-## Capture statistics for SCALAR functions/UDF
+## Propagate statistics for SCALAR functions
 
 Proposers
 
-* Anant Aneja (@aaneja)
 * Prashant Sharma (@ScrapCodes)
+* Anant Aneja (@aaneja)
 
 ## [Related Issues]
 
@@ -13,34 +13,19 @@ Proposers
 2. https://github.com/prestodb/presto/issues/19356
 
 
-Related issues may include GitHub issues, PRs or other RFCs.
-
 ## Summary
+Provide a set of annotations for the scalar functions to specify how input statistics are transformed due to application of the function.
 
-Provide an easy and backwards compatible way for the Presto SCALAR functions(UDFs) to configure a callback or provide certain params via annotation
-to compute how a function impacts source statistics (.e.g distinct value count, null fraction etc...). SCALAR functions have certain 
-characteristics (e.g. they do not impact output row counts) which can be used to provide more options for stats calculation than just
-registering callback.
+A SCALAR function such as `upper(COLUMN1)` could be annotated with `propagateStatistics: true`, and would reuse source statistics of `COLUMN1`.
+Another example : `is_null(COLUMN1)` could be annotated with `distinctValueCount: 2`, `nullFraction: 0.0` and `avgRowSize: 1.0` and would have these stats values hardcoded
 
-A SCALAR function `upper(COLUMN1)` can be optimized if it could be annotated as `propagateStatistics: true` i.e. just reuse source statistics of `COLUMN1`.
-Another example is `is_null(COLUMN1)` can be optimized if it could be annotated as `distinctValueCount: 2`, `nullFraction: 0.0` and `avgRowSize: 1.0`.
-
-We can add a [special case]([ScalarStatsCalculator](https://github.com/prestodb/presto/blob/0.288-edge7/presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsCalculator.java#L118))
-for each example as above or come up with something more general, the current codebase adds special case for functions `CAST(...)`, `is_null(Slice)`, and all arithmetic operators.
-This proposal discusses alternatives to adding special cases or heuristics for optimising scalar udfs or functions.
-
-If none of the parameters can fit a certain functions characteristic then it can use a user provided support function(or a callback) for determining
-function statistics.
-
-Finally, exact statistics cannot be known in advance e.g. the `distinctValueCount` for column will be different after applying `upper(Col1)` function,
-and it depends on the data. A closer estimate can be arrived at - by either using a stochastic approach or sampling.
 
 ## Background
+Currently, we compute new stats for only a handful of functions (`CAST(...)`, `is_null(Slice)`, arithmetic operators) by using a hand rolled implementation in [`ScalarStatsCalculator`](https://github.com/prestodb/presto/blob/0.288-edge7/presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsCalculator.java#L118))
 
-The Presto optimizer cannot estimate statistics for UDFs since it is unaware of how source stats need to be transformed.
+For other functions the optimizer cannot estimate statistics since it is unaware of how source stats need to be transformed.
 [Related Issues](#related-issues) [1] and [2] are examples of queries we could have been optimized better if there was some way to 
-derive source stats. The current implementation simply applies 
-`?` or `VariableStatsEstimate.unknown()` i.e. unknown values for the PlanNode e.g.: 
+derive source stats. `VariableStatsEstimate.unknown()` is propgated for these stats (Shown below as `?` )
 
 ```
 presto:tpcds_sf1_parquet> explain select 1 FROM customer, customer_address WHERE c_birth_country = upper(ca_country);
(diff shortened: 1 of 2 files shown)


RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,420 @@
...
+    /** Use the max value across all arguments to derive the new stats value */
+    USE_MAX_ARGUMENT,
+    /** Sum the stats value of all arguments to derive the new stats value */
+    SUM_ARGUMENTS,
+    /** Propagate the source stats as-is */
+    USE_SOURCE_STATS,
+    /** Use the value of output row count. */
+    ROW_COUNT,
+    /** Use the value of row_count * (1 - null_fraction). */
+    ROW_COUNT_TIMES_INV_NULL_FRACTION,
```
ZacBlanco, line 264: consider just renaming to `NON_NULL_ROW_COUNT`? This name is a bit confusing

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,420 @@
...
+
+Examples of using above annotations for various function in `StringFunctions.java`
+
+1. upper(Slice)
+```java
+    @Description("converts the string to upper case")
+    @ScalarFunction
+    @LiteralParameters("x")
+    @SqlType("varchar(x)")
+    public static Slice upper(@ScalarPropagateSourceStats(propagateAllStats = true) @SqlType("varchar(x)") Slice slice)
```
ZacBlanco, line 282: for readability maybe split annotations onto separate lines?

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,420 @@
...
+3. concat(Slice, Slice): 
+
+```java
+   @Description("concatenates given character strings")
+    @ScalarFunction
+    @LiteralParameters({"x", "y", "u"})
+    @Constraint(variable = "u", expression = "x + y")
+    @SqlType("char(u)")
+    public static Slice concat(@LiteralParameter("x") Long x,
+            @ScalarPropagateSourceStats(propagateAllStats = true, nullFraction = USE_MAX_ARGUMENT, avgRowSize = SUM_ARGUMENTS, distinctValuesCount = ROW_COUNT) @SqlType("char(x)") Slice left,
```
ZacBlanco, line 315: same comment on splitting lines

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,420 @@
...
+    { //...
+}
+```
+ __Propagate all stats but, for nullFraction, avgRowSize and distinctValuesCount follow as specified.__
+
+4. An example of using both `ScalarFunctionConstantStats` and `ScalarPropagateSourceStats`.
+```java
+    @ScalarFunctionConstantStats(minValue = 0)
+    public static long levenshteinDistance(
+            @ScalarPropagateSourceStats(propagateAllStats = true, maxValue = MAX_TYPE_WIDTH_VARCHAR, distinctValuesCount = MAX_TYPE_WIDTH_VARCHAR) @SqlType("varchar(x)") Slice left,
```
ZacBlanco, line 326: Are there many cases where `propagateAllStats` will be false when this annotation is specified? Should we default it to `true` so that it doesn't have to be specified?

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,420 @@
...
+            @SqlType("varchar(y)") Slice right){ 
+    //... 
+  }
+```
+
+* __Phase 2__. Annotate some of the builtin scalar functions with above annotation.
```
ZacBlanco, line 332: I think it would be beneficial to list functions here

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,420 @@
...
+approaches as one could compute via statistical process for those functions user provided stats computation is unavailable. Sampling and other stochastic processes are beyond the
+scope for this RFC.
+
+## Adoption Plan
+
+- What impact (if any) will there be on existing users? Are there any new session parameters, configurations, SPI updates, client API updates, or SQL grammar?
+
+
+Functions without this annotation will not be affect, to remain backwards compatible those functions without this Annotation will run as is. Also, a new session 
+config will be introduced i.e. `optimizer.scalar-function-stats-propagation-enabled=false`, by setting this flag this feature can be turned off.
```
ZacBlanco, line 396: `optimizer.scalar-function-stats-propagation-enabled=false` as currently written is in the form of a "FeatureConfig" (most session properties have only `_` in their name, `-` are not allowed).

```suggestion
config will be introduced i.e. `optimizer_scalar_function_stats_propagation_enabled`, by setting this flag to `false` this feature can be turned off.
```

ZacBlanco (commented):
Should we include information about how we intend to support functions within native worker - or at least show we considered it?

RFC-0004-functions-stats.md
ScrapCodes: There a good number of them, but this is something I am going to keep in mind. If there are more functions with propagateAllStats = true, then make it the default value.

RFC-0004-functions-stats.md
ScrapCodes: I was intending feature config. Session config is specified at https://github.com/prestodb/rfcs/pull/14/files#diff-563baec9193d2d661febe4c2fd74c485dfcb16a9027ee8284dea76be7c1a1915R473 .

RFC-0004-functions-stats.md
ScrapCodes: I intended to go through all the builtin functions and add annotations where it makes sense. List may be huge !

RFC-0004-functions-stats.md
ZacBlanco: Then in the previous line we should call it feature config and not a session config

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,473 @@
...
+How can we measure the impact of this feature?
+
+1. Existing TPCDS and TPCH queries which use functions or `LIKE %STR` in join condition will benefit from this optimization.
+2. Join order benchmark: Reference [2](#references)
+
+## Other Approaches Considered
+
+Based on the discussion, this may need to be updated with feedback from reviewers.
+
+### Approach 1 Trying to find a heuristic based mechanism to estimate functions cost, for example : A function takes a single argument of type `varchar(20)` and
```
aaneja, line 416: This and `Approach 2` are very long headings. Did you mean to add a line break after `Approach 1` ?

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,473 @@
...
+
+#### Pros
+
+* Minimum impact to the user and codebase. Since this feature can be 
+enabled/disabled using a session flag : `scalar_function_stats_propagation_enabled`
+
+#### Cons
+* Problem with heuristic based approach is unknown nature of functions, and thus it can be difficult to come up with heuristics that are always accurate.
+
+### Approach 2: Reference [1. Monsoon](#references). discusses various stochastic approaches to solve the same problem, a user provided stats calculator will complement such
```
aaneja, line 429: nit, replace `1. Monsoon` with just `Monsoon`

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,472 @@
...
+```
+
+#### How does a C++ worker communicate functions stats behavior and constant stats to the coordinator.
+
+The ongoing work described in [RFC-0003](RFC-0003-native-spi.md), with function registries already includes function metadata as field .
+Since we are introducing fields inside the
+[FunctionMetadata.h](https://github.com/facebookincubator/velox/blob/0adc62e3aaf647596ab476f6300a4165e60f6b91/velox/expression/FunctionMetadata.h)
+the new fields have to be sent as is for further processing to the coordinator.
+
+In both Java and C++, builtin functions have extra metadata defined in FunctionMetadata class, which is used while computing estimated stats
```
aditi-pandit, line 364: The side-car returns https://github.com/prestodb/presto/blob/master/presto-function-namespace-managers/src/main/java/com/facebook/presto/functionNamespace/json/JsonBasedUdfFunctionMetadata.java for all Velox functions to the co-ordinator. We would need to enhance this as well for the new Stats. Please add to this RFC.

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,472 @@
...
+
+1. [Monsoon: Multi-step optimization and execution of queries with partially obscured predicates](https://scholar.google.com/citations?view_op=view_citation&hl=en&user=2AjuufAAAAAJ&citation_for_view=2AjuufAAAAAJ:UeHWp8X0CEIC)
+2. "How Good Are Query Optimizers, Really?"
+   by Viktor Leis, Andrey Gubichev, Atans Mirchev, Peter Boncz, Alfons Kemper, Thomas Neumann
+   PVLDB Volume 9, No. 3, 2015
+   http://www.vldb.org/pvldb/vol9/p204-leis.pdf
+
+## Appendix
+
+1. List of functions: Note: At this stage an exhaustive list is not ready (WIP).
```
aditi-pandit, line 467: Please write the Note after specifying all the functions you intend to change.

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,472 @@
...
+                    nullFraction = USE_SOURCE_STATS,
+                    distinctValuesCount = "value(arg_length) <3 ? estimate_ndv(x, value(arg_length), BASIC_LATIN, min(output_row_count, source_stats_ndv())) * (1- source_stats_null_fraction()) : UNKNOWN",
+                    avgRowSize = "value(arg_length)" ) @SqlType("varchar(x)") Slice utf8,
+            @SqlType(StandardTypes.BIGINT) long start,
+            @SqlType(StandardTypes.BIGINT) long length)
+    { //...
+    }
+```
+
+`substr` is at the heart of all `LIKE %X` statements, which is widely used. Without a more powerful framework support it is difficult to support a stats estimation.
```
aditi-pandit, line 396: LIKE in Prestissimo doesn't use substr. It has its own implementation. Please validate that we will pick the correct stats if the implementations of functions are different between C++ and Java.

aditi-pandit (commented):
Thanks @ScrapCodes

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,472 @@
...
+Examples of using above annotations for various function in `StringFunctions.java`
+
+1. upper(Slice)
+```java
+    @Description("converts the string to upper case")
+    @ScalarFunction
+    @LiteralParameters("x")
+    @SqlType("varchar(x)")
+    public static Slice upper(
+            @ScalarPropagateSourceStats @SqlType("varchar(x)") Slice slice)
```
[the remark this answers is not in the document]
ScrapCodes: The default value for propagateAllStats is true, so all the input source stats will be propagated. Will clarify this.

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,472 @@
...
+
+### Goals
+
+1. Provide a mechanism for transforming source statistics for scalar functions
+2. Show impact as plan changes for benchmark queries
+
+### Non-goals
+1. Non-SCALAR functions will not be annotated
+2. Runtime metrics to measure latency/CPU impact of performing these transformations
+3. Give accurate stats by either sampling or applying a statistical technique.
```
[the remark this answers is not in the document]
ScrapCodes: While annotating, our goal is not providing accurate stats always, but the best effort based on function characteristics. In other words, a discussion on more accurate stats technique such as sampling/ etc... is non-goal for this RFC.

RFC-0004-functions-stats.md
ScrapCodes: removed the statement.

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,472 @@
...
+## [Related Issues]
+
+1. https://github.com/prestodb/presto/issues/20513
+2. https://github.com/prestodb/presto/issues/19356
+
+
+## Summary
+Provide a set of annotations for the scalar functions to specify how input statistics are transformed due to application of the function.
+
+A function is just a “black box” for the database system and thus the function may be executed much less efficiently than they could be.
```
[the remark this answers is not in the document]
ScrapCodes: Added a clarification in Background section. Let me know if it looks good.

ScrapCodes:
Thanks @aditi-pandit , @ZacBlanco and @aaneja !

RFC-0004-functions-stats.md
```diff
@@ -0,0 +1,561 @@
...
+stats. 
+
+2. starts_with(Slice, Slice): an example of constant stats.
+```java
+    @SqlNullable
+    @Description("Returns whether the first string starts with the second")
+    @ScalarFunction("starts_with")
+    @LiteralParameters({"x", "y"})
+    @SqlType(StandardTypes.BOOLEAN)
+    @ScalarFunctionConstantStats(distinctValuesCount = 2, minValue = 0, maxValue = 1)
```
feilong-liu, line 236: "minValue = 0, maxValue = 1"
Not sure about this part for boolean value

ScrapCodes: Hi @feilong-liu , thanks for taking a look. 
Yes, analyze on table with boolean column created under HIVE catalog shows NULL for min and max. So will fix this. 

P.S. You can checkout annotation for other functions on our POC PR: https://github.com/prestodb/presto/pull/22974

aaneja (approved):

ZacBlanco (approved):

ScrapCodes:
Hi @feilong-liu , is it good to go?

feilong-liu (approved):

RFC-0005-functions-stats.md
```diff
@@ -0,0 +1,561 @@
...
+
+## Appendix
+
+### List of functions:
+ 1. All functions in presto-main/src/main/java/com/facebook/presto/operator/scalar/MathFunctions.java
+ 2. All functions in presto-main/src/main/java/com/facebook/presto/operator/scalar/StringFunctions.java
+ 3. `getHash` in presto-main/src/main/java/com/facebook/presto/operator/scalar/CombineHashFunction.java
+ 4. `yearFromTimestamp` in presto-main/src/main/java/com/facebook/presto/operator/scalar/DateTimeFunctions.java
+    
+ Note: At this stage an exhaustive list is not ready (WIP).
```
tdcmeehan, line 561: Remove this please

RFC-0005-functions-stats.md
```diff
@@ -0,0 +1,561 @@
...
+Reference [Monsoon](#references), discusses various stochastic approaches to solve the same problem, a user provided stats calculator will complement such
+approaches as one could compute via statistical process for those functions user provided stats computation is unavailable. Sampling and other stochastic processes are beyond the
+scope for this RFC.
+
+## Adoption Plan
+
+- What impact (if any) will there be on existing users? Are there any new session parameters, configurations, SPI updates, client API updates, or SQL grammar?
+
+
+Functions without this annotation will not be affect, to remain backwards compatible those functions without this Annotation will run as is. A new session flag,
```
tdcmeehan, line 525:
```suggestion
Functions without this annotation will not be affected, to remain backwards compatible those functions without this Annotation will run as is. A new session flag,
```

RFC-0005-functions-stats.md
```diff
@@ -0,0 +1,561 @@
...
+                    nullFraction = USE_SOURCE_STATS,
+                    distinctValuesCount = "value(arg_length) <3 ? estimate_ndv(x, value(arg_length), BASIC_LATIN, min(output_row_count, source_stats_ndv())) * (1- source_stats_null_fraction()) : UNKNOWN",
+                    avgRowSize = "value(arg_length)" ) @SqlType("varchar(x)") Slice utf8,
+            @SqlType(StandardTypes.BIGINT) long start,
+            @SqlType(StandardTypes.BIGINT) long length)
+    { //...
+    }
+```
+
+#### virtual function
```
tdcmeehan, line 486: It would be helpful to discuss changes required in Velox with an issue or discussion in the Velox repo.

tdcmeehan (commented):
Just some nits

PR #16, opened 2024-07-18
Add RFC for is_private_ip function

mattcalder-meta:


RFC-0004-isprivateip-udf.md
```diff
@@ -0,0 +1,42 @@
...
+2. Implement a function called `is_private_ip`. The function will accept a single argument of type `IPADDRESS` that will return `true`
+if the argument is a private IP address and `false` otherwise. The actual checks can be completed in two ways:
+    * Some IPv4 private ranges align nicely with dotted-decimal string representation and it is more efficient to use a `STARTS_WITH` than to check if it is equal to a private `IPPREFIX`.
+    * For other private ranges, we'll need to cast to an `IPPREFIX` and see if it equals one of the private ranges.
+3. Add a new test class `presto-main/src/test/java/com/facebook/presto/operator/scalar/sql/TestIPSqlFunctions.java`.
+
+
+## Other Approaches Considered
+
+This function is more efficient to implement on integer representations of IP addresses but we lack of 128-bit integers.
```
elharo, line 33: but we lack of --> , but we lack

elharo (commented):
This feels idiosyncratic to me. I think the broader question to address is whether Presto should or should not have a library of functions for working with IP addresses. I get a little antsy thinking about adding and supporting single functions to cover one user's use cases. 

Anyone know how other SQL systems manage IP address data? 

And if we have IP address functions what about URLs and domain names?

mattcalder-meta:
> This feels idiosyncratic to me. 
> I think the broader question to address is whether Presto should or should not have a library of functions for working with IP addresses. I get a little antsy thinking about adding and supporting single functions to cover one user's use cases. 

Many sizable organization that have a presence on the Internet need to perform routine network analytics for auditing, security (e.g., intrusion & DDoS detection), traffic engineering, performance, and capacity planning. I've done this type of work now in three different companies — in all of them, we used SQL tools to help. Unfortunately, many were not OSS.

* Python, which is what many analytics folks end up using when there a lack of support in SQL, [has this function](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_private). 
* In the Java world this is done with libraries like [IPAddress](https://github.com/seancfoley/IPAddress) and [Google's Guava](https://github.com/google/guava/tree/master/guava/src/com/google/common/net).

> And if we have IP address functions what about URLs and domain names?

Actually, we do have existing functions for [IP addresses](https://prestodb.io/docs/current/functions/ip.html) and [URLs](https://prestodb.io/docs/current/functions/url.html). We also have the `IPADDRESS` and `IPPREFIX` types which help to enable richer functionality.

> Anyone know how other SQL systems manage IP address data?
Far from an exhaustive list but here are the ones I've used before.
* [Clickhouse](https://clickhouse.com/docs/en/sql-reference/functions/ip-address-functions) may be the most relevant.
* [MySQL](https://dev.mysql.com/doc/refman/8.4/en/miscellaneous-functions.html#function_inet-aton)
* [Postgres](https://www.postgresql.org/docs/9.3/functions-net.html)

RFC-0004-isprivateip-udf.md
```diff
@@ -0,0 +1,42 @@
+# **RFC0 for Presto**
+
+## Function to check if IPADDRESS is private
+
+Proposers
+* Matt Calder (mattcalder-meta)
+
+## Summary
+
+Create `is_private_ip` UDF that checks whether an IP address is within private or reserved range that is not globally reachable. We take our definitions from what IANA considers not "globally reachable" in
```
elharo, line 10: The proposal seems to be for a built-in function, not a user defined function.

elharo (commented):
Architecture WG is planning to discuss the meta question of what functions belong in core and which don't and why. We've been a bit slip shod on this.

kaikalur:
So I think we should separate the issue of where functions go etc and let the current process continue and if/when we have a  decision we can refactor things as needed. Let's not block work from being done.

elharo:
It's not quite that easy. One open question is what belongs in core and what belongs in plugins and user defined functions. If we add something to core now, it's really hard to take it out in the future so I'm inclined to be conservative about what we accept. 

I'd also like to make sure that if we do decide to add IP address functions and/or other network related functions to core, that they're consistent with each other for a good developer experience. Ideally they'll also be as consistent as reasonably possible with similar functionality in MySQL, Spark, etc. Taking functions one at a time risks accepting things that look good in isolation but are a confusing mishmash when we put them all together.

We plan to discuss this at today's (Wednesday 7/24/24) Presto architecture working group zoom meeting at 1:00 PM EDT/10:00 AM PDT. Calendar invite with Zoom link in the slack general channel. I encourage everyone to attend.

PR #18, opened 2024-07-20
RFC for IP prefix aggregation function

mattcalder-meta:


elharo (commented):
I feel there's a meta issue between this and #16 about how we do or don't go about addressing IP address types in Presto SQL. I'd prefer not to handle these ona  case by case basis, but rather take a holistic view of address data.

PR #20, opened 2024-07-26
[DRAFT] Grammar for plan constraints

aaneja:


RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,77 @@
...
+
+cardinalityConstraint : CARD (joinNode cardinality)
+
+distributionType : [P]
+   | [R]
+
+joinType : JOIN	(defaults to inner join)
+| IJ
+| LOJ
+| ROJ
```
aaneja, line 50: The first PR for parsing this grammar will not support LOJ and ROJ

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Grammar
+
+```
+planConstraintString : /*! planConstraint [, ...] */
+
+planConstraint : joinConstraint
+| cardinalityConstraint
+
+joinConstraint : joinType (joinNode) [distributionType]
```
ZacBlanco, line 40: Wondering if you considered a slightly different grammar on the constraint where you stick to a single prefix/infix/postfix notation of the join "metadata" parameters, rather than splitting the join type and distribution type on separate sides of the actual join arguments

When I read some of the examples from left to right, I feel it's initially hard for me to parse. Keeping the "metadata" together might be more intuitive to parse in my opinion. It could also very well be my lack of focus on this area which contributes to my confusion though :) 

Example

```
IJ (LOJ (a b) [R]) c
```

Reading left to right out loud (and in my head) I hear:
"Inner Join the result of a left outer join of a and b which is a replicated join with table c"

versus a slightly different notation where you might define a constraint as 
```
joinConstraint : [distributionType] joinType (joinNode) 
```

The previous example now becomes
```
IJ ([R] LOJ (a b)) c
```

"Inner Join the result of a replicated left outer join of a and b with table c"

The second one feels a little easier for me to parse when the "metadata" about the join is in one location before the inputs. However, I think you could also do a postfix or infix version. Prefix felt easier for me at first though.

ZacBlanco: I thought about this some more, and I think the reasoning behind leaning towards the prefix notation in my examples comes from the fact than in English grammar, you will generally describe things as `<adjective> (<adjective>)* <noun>`, so I think the readability of the prefix notation is easier to parse in my head because I see the `distributionType` and `joinType` as "adjectives" which are describing the join. Granted [there are exceptions to that rule](https://www.ucl.ac.uk/internet-grammar/adjectiv/postpos.htm). I could also be convinced that a postfix notation is more suitable, as we don't have to design the constraint language to be readable in English. Just some thoughts

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,77 @@
...
+1. Inner Join constraints - 
+    1. `join (a (b c))` - Join the relations a,b,c as a right deep tree (denoted by the brackets). Use regular rules for determining join distribution
+    2. `join (((a c) [R] b) [P])` - In addition to the join order, use a  REPLICATED `[R]`  join for sub-plan `(a c)`  and PARTITIONED `[P]`  for `(a c) b`
+    3. If an inner join condition does not exist between nodes a CrossJoin is automatically inferred
+2. Cardinality constraints -
+    1. `card (c 10)` - Set the output row count estimate of `c` to `10` 
+    2. `card ((c o) 10)` - When considering a join node of shape `(c o)` set the output row count estimate to `10`
+
+### Other points of note
+- Relation names loosely resolve to WITH query aliases (CTE definitions) and table names.A detailed description of the name resolution is out of scope of this RFC (this will be covered in the implementation PR description)
```
ZacBlanco, line 77: It might be worth fleshing out in more detail exactly how we will deal with name resolution and name scoping for constraints. I'm mainly concerned about name clashes between CTEs/subqueries/aliases (I'm not sure of the exact terminology).

Consider the following simple example:

```sql
WITH t as (
    SELECT * FROM
    (VALUES 1, 2) as t(a)
    LIMIT 1)
SELECT * FROM t;
```

How would you write a constraint which would be able to unambiguously reference the `t` from `WITH t` and the `t` from `VALUES ... t(a)`? This is important because each table `t` has a different cardinality `VALUES t` has a cardinality of 2, while `WITH t` has cardinality `1` due to the `LIMIT`. I would imagine we don't want to limit our ability to define constraints when names clash.

aaneja: >I would imagine we don't want to limit our ability to define constraints when names clash.
In the first implementation, we will require users to define unambiguous names for cases like this, i.e they would have to rewrite their SQL.
However, this can improved upon. We can build the name matcher to allow us to specify nested names like `t.t`.
Since the name matching is implementation dependent, and does not require a grammar rule, I chose to skip describing it here. I will be covering it in the implementation PR

RFC-0007-plan-constraints-grammar.md
aaneja: - IMO the join type is a "verb" and the table names are "nouns", so `(a IJ b)` is read as `a INNER JOIN b`. The properties `R` and `P` are extra attributes of the join, so they can be skipped or specified after the "join sentence" is made

- For outer joins, it is easier to understand what is the outer-side if it follows the SQL convention, so `(a LOJ b)` is well understood as `a LEFT OUTER JOIN b`. This also makes more of a difference when nesting goes beyond a few tables, e.g consider -
```
In order : (((1 LOJ 2) ROJ 3) LOJ 4)
```
vs
```
Prefix : (LOJ (ROJ (LOJ 1 2) 3) 4)
```

ZacBlanco: I can see how using infix makes seeing the left/right-hand sides of the join easier. In that case, should we update the `joinNode` definition to use infix notation for the join type? The current spec has `joinType` being specified before the relation name.

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+
+### Examples of constraints
+
+1. Inner Join constraints - 
+    1. `join (a (b c))` - Join the relations a,b,c as a right deep tree (denoted by the brackets). Use regular rules for determining join distribution
+    2. `join (((a c) [R] b) [P])` - In addition to the join order, use a  REPLICATED `[R]`  join for sub-plan `(a c)`  and PARTITIONED `[P]`  for `(a c) b`
+    3. If an inner join condition does not exist between nodes a CrossJoin is automatically inferred
+2. Cardinality constraints -
```
aditi-pandit, line 72: Can you give an example that has both join and cardinality constraints ?

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,77 @@
...
+
+| (relationName joinNode)
+
+| (joinNode relationName)
+
+| (joinNode joinNode [, ...])
+```
+
+
+### Examples of constraints
```
aditi-pandit, line 66: Would be great if the example showed the outer SQL snippet as well.

RFC-0007-plan-constraints-grammar.md
aditi-pandit: I feel that we always see distribution as the last piece of a table string in EXPLAIN plans. So the current syntax looks alright for it. I'm struggling more with LOJ being prefix instead of infix. Any particular reason for going with prefix ?

RFC-0007-plan-constraints-grammar.md
aditi-pandit: Is there a case to refer to something internal to a with clause in the outer context ? Generally, I would imagine constaints to be placed within the with definition itself.

Using an outer context is more complicated since there could be multiple uses of the CTE and so we would have to think about disambiguating the uses of the CTE as well.

RFC-0007-plan-constraints-grammar.md
aaneja: Yes, I think the grammar here is buggy. @ClarenceThreepwood can you fix up the grammar to reflect our understanding here. I think infix representation is easiest to understand, and we should start with it. The grammar can be expanded at a later stage to allow for terser specification.

RFC-0007-plan-constraints-grammar.md
aaneja: Added

RFC-0007-plan-constraints-grammar.md
aaneja: @ClarenceThreepwood  I took a stab at making the grammar more precise (and closer to parseable ANTLR). Please let me know if this is in-line with what you had in mind

ZacBlanco (approved):

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,123 @@
...
+    1. `card (c 10)` - Set the output row count estimate of `c` to `10` 
+    2. `card ((c o) 10)` - When/If considering a join node of shape `(c o)` set the output row count estimate to `10`
+3. Both type of constraints -
+    `join (c o) card ((c o) 10)` - Force a join-sub-graph of nodes `c InnerJoin o`. For this join node, set the output row count estimate to `10`
+
+#### Full SQL examples of queries with constraints
+
+Force the join of `c` and `cte` which is otherwise ignored, see [19354](https://github.com/prestodb/presto/issues/19354) 
+```
+/*! join ((c cte) n) */
```
aditi-pandit, line 97: Nice

aditi-pandit (approved):
Thanks @aaneja

RFC-0007-plan-constraints-grammar.md
aaneja: I've updated the grammar. It parses correctly in my local testing

aaneja:
With #[3fcc611](https://github.com/prestodb/rfcs/pull/20/commits/3fcc6118641b15477ae82bc2b069b926f86cfe23) I've update the grammar to make the join specification more explicit.

This disallows specifying grouped relations like `(A B C)` . Instead we need to specify the full join order like `((A B) C)`. This will makes specifying ambiguous group relations like `(A LOJ B ROJ C)` impossible

At a later stage, we can enhance the grammar so that inner join only groups like `(A B C)` can also be specified

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,160 @@
...
+
+planConstraintString : CONSTRAINTS_START_MARKER (planConstraint)* CONSTRAINTS_END_MARKER;
+
+
+```
+
+
+### Examples of constraints
+
+1. Inner Join constraints -
```
jaystarshot, line 120: Why can't we just use replicated or paritioned syntax? Imo the brackets and` join ((a c [R]) b [P])` might get complicated quickly

aaneja: Can you elaborate on what this would look like for the cited example ?

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,160 @@
+# **RFC007 for Presto**
+
+## Grammar for Plan Constraints
```
rschlussel, line 3: can you add an overview of other engines that allow query hints and what their grammar looks like?

RFC-0007-plan-constraints-grammar.md
jaystarshot: Like [dremio](https://docs.dremio.com/cloud/reference/sql/commands/SELECT/#distributing-data-evenly-across-execution-engines-during-joins) or [databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-hints.html#join-hints).

Current way also is fine but suggestion is to use explicit names for broadcast/partition etc since the goal of this is to allow users to explictly set the types

RFC-0007-plan-constraints-grammar.md
aaneja: The [overview doc](https://prestodb.io/wp-content/uploads/Search-Space-Improvements-Plan-Constraints.pdf) links out to how other engines implement plan constraints. I copied over the relevant info to the background section

RFC-0007-plan-constraints-grammar.md
aaneja: I'm not opposed to this; my only gripe with it is that it that the plan constraint string can get quite verbose, e.g for a 4-table join order - ` join (d ((a c [REPLICATED]) b [PARTITIONED]) [PARTITIONED] )`

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,168 @@
...
+LOJ        : 'LOJ' ;
+ROJ        : 'ROJ' ;
+IJ         : 'IJ' ;
+LPAREN     : '(' ;
+RPAREN     : ')' ;
+JOIN_DIST_PARTITIONED : '[P]' ;
+JOIN_DIST_REPLICATED : '[R]' ;
+CARD : 'CARD' ;
+JOIN : 'JOIN' ;
+CONSTRAINTS_START_MARKER :'/*!';
```
kaikalur, line 57: Spark uses /*+..*/ syntax so msybe best

kaikalur (approved):

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,168 @@
...
+
+This document proposes a grammar for specifying plan constraints
+
+## Background
+
+Plan constraints can be used to lock down critical aspects of an execution plan, such as access method, join method, and join order.
+
+Most commercially available cost-based optimizers support some form by which a user could express plan constraints that enable these two use-cases. Common terms used by DB vendors to describe this set of features include "plan constraints", "optimization guidelines" or "plan hints".
+
+There are two broad approaches as to how these plan constraints are stored and represented. Vendors such as [Oracle](https://docs.oracle.com/cd/B10500_01/server.920/a96533/hintsref.htm) & [Vertica](https://www.vertica.com/docs/10.0.x/HTML/Content/Authoring/SQLReferenceManual/LanguageElements/Hints/Hints.htm) provide a grammar/syntax where the plan constraint may be expressed inline with the sql query. The parser then picks up the constraint, validates it, and then passes it on to the query optimizer.
```
rschlussel, line 25: It would helpful to describe the syntax of other systems and how this one compares. like if we're doing something different from everyone else we need a very clear reason why.

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,168 @@
...
+    ;
+
+joinType
+    : LOJ
+    | ROJ
+    | IJ
+    ;
+
+joinAttribute
+    : JOIN_DIST_PARTITIONED
```
rschlussel, line 97: just call this PARTITIONED.  JOIN_DIST_PARTITIONED is clunky to write and to remember

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,168 @@
...
+    | JOIN_DIST_REPLICATED
```
rschlussel, line 98: call this BROADCAST, which will be much more familiar to users.

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,168 @@
+# **RFC007 for Presto**
+
+## Grammar for Plan Constraints
```
rschlussel, line 3: I would call this feature Plan Hints rather than Plan Constraints. The optimizer may not follow the hints (though we should issue a warning if we don't)

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,168 @@
...
+    : LPAREN joinedRelation RPAREN
+    ;
+
+joinTypeOrDefault
+    : joinType
+    | { "IJ" } // Default value
+    ;
+
+joinType
+    : LOJ
```
rschlussel, line 91: why is this needed?

kaikalur:
A high level comment - plan level hints make them not accessible to most users. We should keep them more high level like:

```
BROADCAST
MAPJOIN
HIGHDISTINCT
LOWDISTINCT 
RANDOM_SHUFFLE
FORCE_SHUFFLE
SHARDING
PAYLOAD
SALTING
...
```

that tell the planner something about the data shape. Planner is a not concept that most users understand. But some of the others we can easily explain to users. Also things fragments can change based on other options. So not sure it's good to use those.

RFC-0007-plan-constraints-grammar.md
rschlussel: Do we need this fine grained kind of control.  I like the simplicity of just specifying the partitioning of the table, and every other example I found seems to take that approach. 
1. [Spark](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-hints.html#join-hints): `BROADCAST(T1)`
2. [Oracle](https://docs.oracle.com/cd/B10500_01/server.920/a96533/hintsref.htm#5544): `PQ_DISTRIBUTE(s BROADCAST, NONE)` (which I also find very clunky).
3. [Vertica](https://www.vertica.com/docs/10.0.x/HTML/Content/Authoring/SQLReferenceManual/LanguageElements/Hints/distrib.htm): you specify inline distrib(L,B)
4. [Dremio](https://docs.dremio.com/cloud/reference/sql/commands/SELECT/#distributing-data-evenly-across-execution-engines-during-joins): mark `BROADCAST` inline

You can specify to use syntactic join order if you want to control the join ordering in a more complex way.

I find the existing syntax complex and hard to use.

RFC-0007-plan-constraints-grammar.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+### Examples of constraints
+
+1. Inner Join constraints - 
+    1. `join (a (b c))` - Join the relations a,b,c as a right deep tree (denoted by the brackets). Use regular rules for determining join distribution
+    2. `join ((a c [R]) b [P])` - In addition to the join order, use a  REPLICATED `[R]`  join for sub-plan `(a c)`  and PARTITIONED `[P]`  for `(a c) b`
+    3. If an inner join condition does not exist between nodes a CrossJoin is automatically inferred
+2. Cardinality constraints -
+    1. `card (c 10)` - Set the output row count estimate of `c` to `10`
```
rschlussel, line 133: I would rather not expose this kind of control.  Better to specify what you want to have happen with this table, and otherwise leave it be. Seems pretty risky to encode cardinality estimates in the query text.

RFC-0007-plan-constraints-grammar.md
rschlussel, line 57: agree with Sreeni that `/*+` seems more standard.

RFC-0007-plan-constraints-grammar.md
aaneja: >I would call this feature Plan Hints rather than Plan Constraints.  

Sure, will rename
Agreed on providing a warning if some/all of the plan hint is not applied

RFC-0007-plan-constraints-grammar.md
aaneja: I can add a summary section describing the hint syntax, but IMO we can be opinionated about the syntax and make the right tradeoffs for what works best for us.
Happy to discuss this more on a case-by-case basis

RFC-0007-plan-constraints-grammar.md
aaneja: [DB2](https://www.ibm.com/docs/en/db2/11.5?topic=requests-any-join) supports 'join requests' similar to these join hints. It has variants that specify the join method too
Using a join hint allow us to :
1. Specify partial join order - you don't need to force syntactic join on parts on the rest of the query to force a specifc join shape for part of the query tree
2. Hardcode join orders for tools that generate SQL and cannot guarantee syntactic SQL generation

RFC-0007-plan-constraints-grammar.md
aaneja: Can you elaborate on the risk ? I see this `card` estimate as a simple way to fixup estimation errors for when - 
1. Presto doesn't have stats, but the user is aware of maxcard for a base table
2. Presto is unable to figure out the right stats because it has a missing stats rule (e.g see the `WITH CTE` example at L144 below)
3. Provides a good debugging aid for devs to quickly test what-if scenarios wrt to cardinality estimates

RFC-0007-plan-constraints-grammar.md
rschlussel: but the type of join is determined by the query, so even if we support join hints, we shouldn't need to specify if it's an inner or outer join.

RFC-0007-plan-constraints-grammar.md
rschlussel: I agree we can be opinionated and choose what works best for us, but we should still learn from what others do and where we vary, we should understand why we are going a different route/ are consciously choosing it. Being more similar to other systems people may have encountered can also make it more user friendly.

RFC-0007-plan-constraints-grammar.md
rschlussel: The risk I see is that if you are specifying a cardinality of x, you are probably doing it because you want the optimizer to do some particular thing about it. But you don't really know what the optimizer will do with that information, and it could do one thing for a while, and then in a new release there's an optimizer change and it does something else.  Because you aren't directly controlling what happens when you specify the cardinality, it's hard to say how it might affect the query, and could be hard to debug if the performance degrades (vs. if you e.g. specify broadcast join and your data gets bigger, it's very clear what happens)

 It can also get out of sync with the data (there is always a risk with hand tuning that the optimization will no longer be relevan or will perform worse as the data changes, but specifying a specific cardinality estimate can have more varied and unknown effects).

RFC-0007-plan-constraints-grammar.md
rschlussel: Got it.  I see the appeal of specifying a partial join order for auto generated queries, but also would like it to be easy to specify join hints for the common case where people just want to mark some table for broadcast or similar. I wonder if there's an alternative approach we can use to achieve both of these goals

RFC-0007-plan-constraints-grammar.md
aaneja: If/when we support [outer join reordering](https://prestodb.io/wp-content/uploads/Search-Space-Improvements_-Comprehensive-Join-Enumeration.pdf), we would need these to specify hints for outer-join reordering. Of course, if the planner never evaluates these hint choices, these hints are ignored

RFC-0007-plan-constraints-grammar.md
aaneja: We could add Spark style independent hints -
1. `BROADCAST(T1)` - Always broadcast T1 in any join
2. `BROADCAST(a b)` - If the optimizer chooses a logical join of `a` and b`, and this is a sub-tree of another join graph, use BROADCAST for the join distribution

These will be complementary to the join-order syntax. So the below join hints  are equivalent but provide flexibility to the user -
```
BROADCAST(a b)  join(c (ab))
join (c (a b)[R])
```

Will think of more examples and incorporate them

RFC-0007-plan-constraints-grammar.md
aaneja: I agree that this is a powerful knob to give to the users, one that we can document as such (with caveats)  

To aid debugging, we can add `CardEstimateBasedSourceInfo` similar to how we have a `HistoryBasedSourceInfo` that will make it clear to the users how the cardinality was arrived at in EXPLAIN/EXPLAIN ANALYZE (and event listener etc.)

Additional safeguards like warnings & metrics can be incorporated too if the stats estimate differ widely from actual runtime observed cardinality

PR #23, opened 2024-08-01
Add RFC for Multi-Cluster Resource Management 

prithvip:
A proposal for multi-cluster resource management

### DIFF
--- a/RFC-0005-multi-cluster-resource-management.md
+++ b/RFC-0005-multi-cluster-resource-management.md
@@ -1,0 +1,64 @@
+# **RFC5 for Presto**
+
+## [Multi Cluster Resource Management]
+
+Proposers
+
+* Prithviraj Pandian, Meta 
+* Abhisek Saikia, Meta
+
+## Summary
+
+We intend to implement functionality and interfaces in Presto to integrate with an optional external queueing service. 
+
+## Related Issues 
+
+The idea of moving the queueing phase out of the coordinator into a separate process is an old one dating back many years. Here are a few previous proposals/discussions: 
+* https://github.com/prestodb/presto/pull/12176
+* https://github.com/trinodb/trino/issues/391
+* https://github.com/prestodb/presto/issues/10174
+* https://github.com/prestodb/presto/issues/13198
+
+## Background
+
+Currently, each Presto cluster owns its own resource management and queue. The resource manager of each cluster decides which query in the cluster’s queue should be executed next, based on logic described in the cluster’s [resource group configuration](https://prestodb.io/docs/current/admin/resource-groups.html). 
+
+This status quo of cluster-level queueing works fine to manage one or a few clusters, but quickly becomes problematic when managing a deployment of tens or hundreds of Presto clusters. Some issues include: 
+
+1. Due to the queue being fragmented across multiple clusters, and the admission decision being made at the cluster level, resource group admission decisions are not correct globally, although they could be correct locally. For example, it is possible that one cluster might admit a query of lower priority than a query in another cluster’s queue. 
+2. Since query concurrency control is applied on a per-cluster basis, global concurrency control can only be achieved, if the concurrency limit is a multiple of the number of clusters. For example, it is currently impossible to enforce a resource group with a concurrency limit of 1 query, across multiple clusters. 
+3. Load-balancing queries effectively across multiple clusters requires the load balancer to understand the resource group and queue state of each cluster. This is difficult to achieve, and in practice, stale information leads to poor load balancing. 
+4. Since the query is queued on the cluster itself, restarting clusters for maintenance reasons requires either that the queued queries be killed, or a lengthy drain time. 
+
+## Proposed Implementation
+
+We propose to add support in Presto for an optional queueing service that sits upstream of multiple Presto clusters. This queueing service will serve as the global entry-point for all queries from the client and will implement the QueuedStatement protocol. This queueing service is responsible for: 
+
+1. Authenticating the request 
+2. Adding the query to the global queue 
+3. Responding to clients polling for the queueing status of their query
+4. Picking the next query to execute 
+5. Selecting the best cluster for execution of this query
+6. Dispatching this query to the cluster
+7. Redirecting the client to the picked cluster
+
+The internal details of this queueing service are still under development, and this document focuses on the changes required to Presto for such a service to work. If there is interest from the community, we can provide an open-source reference implementation of this queueing service, once mature. 
+
+### Diagram
+![Presto Github Issue GRM](https://github.com/user-attachments/assets/7b2af8fd-5d82-411d-ac97-795e54e35164)
+
+### Work Items
+Because we want this to be transparent to Presto clients, the front-end of the queueing service will reuse the existing functionality of classes in the presto-dispatcher package such as QueuedStatementResource, DispatchManager, DispatchQuery, etc. The high-level work items are:
+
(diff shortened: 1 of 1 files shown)


prithvip:
cc @tdcmeehan @rschlussel @arhimondr @NikhilCollooru

RFC-0005-multi-cluster-resource-management.md
```diff
@@ -0,0 +1,64 @@
...
+
+1. Authenticating the request 
+2. Adding the query to the global queue 
+3. Responding to clients polling for the queueing status of their query
+4. Picking the next query to execute 
+5. Selecting the best cluster for execution of this query
+6. Dispatching this query to the cluster
+7. Redirecting the client to the picked cluster
+
+The internal details of this queueing service are still under development, and this document focuses on the changes required to Presto for such a service to work. If there is interest from the community, we can provide an open-source reference implementation of this queueing service, once mature.
```
jaystarshot, line 45: +1 we at uber will also be interested. What language will this be implemented in? Will it be similar to https://github.com/trinodb/trino-gateway?

prithvip: This will be implemented in Java. It is similar in some ways, but Trino Gateway is only a routing layer, this queueing service will also include queueing and scheduling. We would likely expose an interface to plug in any custom routing logic.

RFC-0005-multi-cluster-resource-management.md
```diff
@@ -0,0 +1,64 @@
...
+3. Responding to clients polling for the queueing status of their query
+4. Picking the next query to execute 
+5. Selecting the best cluster for execution of this query
+6. Dispatching this query to the cluster
+7. Redirecting the client to the picked cluster
+
+The internal details of this queueing service are still under development, and this document focuses on the changes required to Presto for such a service to work. If there is interest from the community, we can provide an open-source reference implementation of this queueing service, once mature. 
+
+### Diagram
+![Presto Github Issue GRM](https://github.com/user-attachments/assets/7b2af8fd-5d82-411d-ac97-795e54e35164)
```
jaystarshot, line 48: You might want to add this diagram as a separate file like [here](https://github.com/prestodb/rfcs/pull/10/files#diff-7c4be749b64ffea6678fccf339b0f7430c613636a3ed4ca5a7a751abc67c71d9)

RFC-0005-multi-cluster-resource-management.md
```diff
@@ -0,0 +1,64 @@
...
+
+1. Allow for the queueing service to forward all required query metadata to the cluster, such as query ID, authenticated identity, query state timings, etc. This will involve changes to QueuedStatementResource, QueryStateMachine, and DispatchManager, so that coordinator can create a query with an already populated QueryId and AuthorizedIdentity.
+2. De-couple the dispatching and execution phases of the query lifecycle. This will involve separating and removing dependencies that are required for execution of the query, but not for queueing, such as Metadata class. We will create a new module, presto-dispatcher, that has all classes required for the queueing phase, such as QueuedStatementResource, DispatchManager, ResourceGroupManager, etc. The presto-main module will have a dependency on this new module. Interfaces such as ResourceGroupManager and DispatchQuery will be in presto-dispatcher, with implementations such as InternalResourceGroupManager and LocalDispatchQuery in presto-main. In addition, it could be necessary to move some classes into a common module, if they are required by both presto-main and presto-dispatcher.
```
jaystarshot, line 54: > We will create a new module, presto-dispatcher
Will presto-dispatcher be used by the new service?

jaystarshot (approved):

RFC-0005-multi-cluster-resource-management.md
```diff
@@ -0,0 +1,64 @@
...
+
+The internal details of this queueing service are still under development, and this document focuses on the changes required to Presto for such a service to work. If there is interest from the community, we can provide an open-source reference implementation of this queueing service, once mature. 
+
+### Diagram
+![Presto Github Issue GRM](https://github.com/user-attachments/assets/7b2af8fd-5d82-411d-ac97-795e54e35164)
+
+### Work Items
+Because we want this to be transparent to Presto clients, the front-end of the queueing service will reuse the existing functionality of classes in the presto-dispatcher package such as QueuedStatementResource, DispatchManager, DispatchQuery, etc. The high-level work items are:
+
+1. Allow for the queueing service to forward all required query metadata to the cluster, such as query ID, authenticated identity, query state timings, etc. This will involve changes to QueuedStatementResource, QueryStateMachine, and DispatchManager, so that coordinator can create a query with an already populated QueryId and AuthorizedIdentity.
```
tdcmeehan, line 53: Can you go into more detail on what sort of changes would be needed in the client protocol (`QueuedStatementResource`)?

RFC-0005-multi-cluster-resource-management.md
tdcmeehan: It seems like this is routing + queueing.  We already have a module for the routing component, that's the `presto-router`.  Could the reference implementation be added there--i.e., could the Presto Router be configured to also take on queueing?

tdcmeehan: Also, I read this as, a separate internal implementation will be developed, but also a reference implementation will separately be developed.  i.e. there will be two implementations.  Could we all work off of the reference implementation, and it can be configured through an SPI to integrate with internal components, rather than have two separate implementations?

tdcmeehan (commented):
Can you add some details on how authentication and authorization will work between the queueing service and the individual coordinators?

RFC-0005-multi-cluster-resource-management.md
prithvip: @tdcmeehan We have to see how the queueing implementation evolves, so this is TBD. We expect to use internal components and its not clear how easily we can configure these through an SPI, while still maintaining usefulness. Two integration points that could make sense are an SPI for ResourceGroupManager and for the router. The reference implementation could use the presto-router. We can discuss these points when the design is more mature. For this RFC, I want to focus on the changes required to integrate with this service. Mostly the work here for now is straightforward refactoring.

RFC-0005-multi-cluster-resource-management.md
prithvip: Yes, the queueing service will have a dependency on presto-dispatcher. The queueing service "implementation" of the dispatcher is as standalone server, and the presto-main "implementation" of the dispatcher will be in-process with the coordinator. The code is already organized this way - for example, DispatchQuery is an interface that would live in presto-dispatcher, with LocalDispatchQuery being the implementation in presto-main and a hypothetical RemoteDispatchQuery that could be implemented by the queueing service.

prithvip:
@tdcmeehan 
> Can you add some details on how authentication and authorization will work between the queueing service and the individual coordinators?

The queueing service will authenticate and, as a result of the authentication, will create an Identity object, same as how the coordinator does it today. This Identity object will be serialized and passed as a HTTP header to the coordinator's QueuedStatementResource. The coordinator will have a trusted relationship with the queueing service, and do authentication to make sure the request is being received from the queueing service (configured trusted principal). This trusted relationship is similar to how Presto Proxy works today with coordinator. There is no change to authorization and authorization requests will happen at the coordinator (during plan time). 

> Can you go into more detail on what sort of changes would be needed in the client protocol (QueuedStatementResource)?

There will be no changes required to any client. QueuedStatementResource will be extended with a new POST endpoint, as "/v1/statement/{queryId}?slug=SLUG". Two headers will be added: 1) forward the serialized Identity object 2) forward query state timings such as waitingForPrerequisitiesTime, queuedTime, etc. DispatchManager will require some minor modifications to accept a externally-minted queryId, slug, and pre-existing state timings. This is the high level overview, but of course, more detailed discussion can happen on the PR.

PR #25, opened 2024-08-14
Add REST API for remote functions RFC

tdcmeehan:


prithvip:
"This RFC proposes to extend the
remote function design to allow for dynamic function registration and execution."

Don't we already support dynamically registered functions today? We can do CREATE FUNCTION and it will create the function in the function namespace manager, which could create it in Metastore.

prithvip:
Why is REST API like GET /v1/functions necessary? Since we can get this information already through "SHOW FUNCTIONS" today?

tdcmeehan:
> Don't we already support dynamically registered functions today? We can do CREATE FUNCTION and it will create the function in the function namespace manager, which could create it in Metastore.

Let me clarify, while we can `CREATE FUNCTION`, this takes a SQL expression.  Our SQL grammer permits the addition of an identifier, but we don't use it, so effectively we can't create an externally defined function (like a JAR or container), it is presumed that this is externally managed.  So the goal is to enable that use case, and also provide a reference implementation that shows how that works end to end.

> Why is REST API like GET /v1/functions necessary? Since we can get this information already through "SHOW FUNCTIONS" today?

The purpose of a REST spec is to allow for a straightfoward implementation of function namespace managers.  Currently, Presto does not have an opinion on the format of the remote function server.  This adds an implementation of a function namespace manager that interacts with this REST spec, a reference implementation of this REST spec that can be used as is or for testing or to enable Java SPI-defined functions for C++ clusters, and integration with Prestissimo so that C++ clusters can make calls to the function server.

RFC-0007-remote-functions.md
```diff
@@ -0,0 +1,219 @@
...
+## [Related Issues]
+
+* https://github.com/prestodb/presto/issues/14053
+
+## Summary
+
+### Dynamic functions in remote function servers
+
+A new REST API is defined, along with a REST plugin implementation, which allows for consistent and unified metadata and execution
+of remote functions in a single API definition.  This API is designed to be extensible, allowing for the definition of new functions
```
aditi-pandit, line 21: Do you mean dynamic here instead of extensible ? Can we presume that new functions can be defined and used without a server restart ?

RFC-0007-remote-functions.md
```diff
@@ -0,0 +1,219 @@
...
+* Deprecate existing function server APIs
+* Provide support for aggregate functions or table-valued functions
+
+## Proposed Implementation
+
+### Design
+
+Fundamentally, the design of the REST API for remote functions will be based on the existing `FunctionNamespaceManager` interface SPI.
+Additionally, for Presto C++, currently there is no corresponding SPI for function namespace managers.  This RFC proposes to
+create a new REST-based implementation of the function execution framework.  By creating a REST API for remote functions that
```
aditi-pandit, line 58: Did you evaluate REST vs gRPC ?

RFC-0007-remote-functions.md
```diff
@@ -0,0 +1,219 @@
...
+unifies the metadata and execution of remote functions, C++ deployments can similarly customize their function servers
+by implementing the REST API, in the same way that current Presto Java users can implement the `FunctionNamespaceManager` SPI.
+
+The REST API will power all `FunctionNamespaceManager` method implementations, including listing functions, retrieving function
+metadata, executing functions, and providing DDL support for functions (`ADD` and `DROP` support).
+
+#### Presto C++ special considerations
+
+The current Presto C++ implementation does not have a `FunctionNamespaceManager` SPI.  This RFC proposes to extend the Velox
```
aditi-pandit, line 67: Are the function servers configured in a static list or is there a dynamic discovery of function servers ? Will the function server need to announce itself to a discovery service ?

RFC-0007-remote-functions.md
```diff
@@ -0,0 +1,219 @@
...
+
+> Endpoint: /v1/functions/{schema}/{functionName}
+>
+> HTTP verb: GET
+>
+> Request body: empty
+>
+> Response body: JSON array of function metadata objects
+
+Returns the complete listing of functions in the specified schema with the specified function name.
```
aditi-pandit, line 127: This should return the function Ids as well so that they can be used in subsequent calls ?

RFC-0007-remote-functions.md
```diff
@@ -0,0 +1,219 @@
...
+
+#### Add a function
+
+> Endpoint: /v1/functions/{schema}/{functionName}
+>
+> HTTP verb: POST
+>
+> Request body: JSON object representing the function to be added
+>
+> Response body: the function ID of the newly created function
```
aditi-pandit, line 137: Would be great to give more details about functionId. functionId is local to the function server. Will Presto use it at all ?

RFC-0007-remote-functions.md
```diff
@@ -0,0 +1,219 @@
...
+>
+> HTTP verb: GET
+>
+> Request body: empty
+>
+> Response body: JSON array of function metadata objects
+
+Returns the complete listing of functions in the specified schema with the specified function name.
+
+#### Add a function
```
aditi-pandit, line 129: How do you envision this being used ? PrestoSQL could expose CREATE FUNCTION that uses this API. But then how do you give access to the function body ? 

How do you make this work across function implementations in different languages ?

RFC-0007-remote-functions.md
```diff
@@ -0,0 +1,219 @@
...
+> Request body: JSON object representing the function to be added
+>
+> Response body: the function ID of the newly created function
+
+Creates a new function in the specified schema with the specified name.  The function object will contain the metadata of the
+function, including its arguments, return type, and other metadata.  It will return an identifier representing this
+specific function, which is useful to differentiate multiple functions which share the same name but have different
+arguments.
+
+#### Update a function
```
aditi-pandit, line 144: What is the e2e flow of interactions in which this API is used ?

RFC-0007-remote-functions.md
```diff
@@ -0,0 +1,219 @@
...
+unifies the metadata and execution of remote functions, C++ deployments can similarly customize their function servers
+by implementing the REST API, in the same way that current Presto Java users can implement the `FunctionNamespaceManager` SPI.
+
+The REST API will power all `FunctionNamespaceManager` method implementations, including listing functions, retrieving function
+metadata, executing functions, and providing DDL support for functions (`ADD` and `DROP` support).
+
+#### Presto C++ special considerations
+
+The current Presto C++ implementation does not have a `FunctionNamespaceManager` SPI.  This RFC proposes to extend the Velox
+function server to support the REST API for remote functions.  This will allow for the definition of functions at runtime in
```
aditi-pandit, line 68: The function server and Presto server should have independent lifetimes. If the user wanted to upgrade their function servers, then how does that proceed ?

aditi-pandit (commented):
@tdcmeehan : Had a bunch of comments.

RFC-0007-remote-functions.md
tdcmeehan: Dynamic might convey the meaning better.  Yes, new functions can be defined and used without a restart with this API.

RFC-0007-remote-functions.md
tdcmeehan: The goal of this RFC is to add an API that is general purpose and which can easily be integrated into one's existing infrastructure.  By deploying Presto, you already deploy a RESTful service, so using gRPC would involve an additional communication protocol.

RFC-0007-remote-functions.md
tdcmeehan: Function ID is used to disambiguate functions with the same name but different signatures.  Two functions with the same name but different signatures are different functions, and the function ID is used to separate them apart.  I will add this to the RFC.

RFC-0007-remote-functions.md
tdcmeehan: This is implicit.  I will update this RFC so it is more clear.

RFC-0007-remote-functions.md
tdcmeehan: Here is one example.  CREATE FUNCTION stores a function definition within the function server.  After this, the version of the function API has changed, which causes a cache invalidation, triggering updated function metadata to be returned.  Suppose we want to use the function.  We now have the signature of the function, thanks to the cache invalidation being triggered above.  A function handle is created and sent to workers that contains the function metadata required to execute the function.  Workers send the page data and function metadata to the server for execution.

This will work across languages because the underlying server may support multiple languages.

RFC-0007-remote-functions.md
tdcmeehan: Because functions are versioned, this is supported in the API.  So long as a functions server faithfully returns functions by their version, then this is supported.

RFC-0007-remote-functions.md
```diff
@@ -0,0 +1,219 @@
...
+
+Deletes the function in the specified schema with the specified name and function ID.
+
+#### Execute a function
+
+> Endpoint: /v1/functions/{schema}/{functionName}/{functionId}/{version}
+>
+> HTTP verb: POST
+>
+> Request body: Presto Page of input data
```
jaystarshot, line 174: Can we also support arrow here?

jaystarshot: Nvm its just bytes

tdcmeehan: I will update the spec to show that Arrow is an acceptable option.

RFC-0007-remote-functions.md
tdcmeehan: Added a section on service discovery

RFC-0007-remote-functions.md
tdcmeehan: Added a flow diagram

RFC-0007-remote-functions.md
tdcmeehan: Added this detail to the doc

RFC-0007-remote-functions.md
tdcmeehan: Added a flow diagram

RFC-0007-remote-functions.md
tdcmeehan: Added a sentence describing initial support for Presto Page but later support for Arrow.

aditi-pandit (approved):
Thanks @tdcmeehan

PR #26, opened 2024-08-19
Add RFC for configurable plan checker

tdcmeehan:


RFC-0008-plan-checker.md
```diff
@@ -0,0 +1,137 @@
...
+class, and the `PlanChecker` class will be modified to run the SPIs during the appropriate phase.
+
+A new endpoint to the Presto sidecar will be added that will attempt to convert a Presto plan to a Velox plan.  In the
+`presto-native-plugin` module, a new implementation of the SPI will be added which will add a check to the `PlanChecker` class
+which will call the sidecar to attempt to convert the plan.  If the conversion fails, the plan checker will fail the plan.  It will
+be added at the `validateFragmentedPlan` phase--this is because it's not until plan fragmentation occurs that we know which portions
+of the plan will be executed in the coordinator, and which will be executed in the workers.
+
+The code for the plan checker will run `PrestoToVeloxQueryPlan`, which is used by the workers to convert the Presto plan fragment
+to a Velox plan fragment.  If the conversion fails, the plan checker will fail the plan, returning with it the reason for the failure.
```
BryanCutler, line 68: It might be obvious, but just to add that since the plan checker is a plugin, it is not run at all if the plugin is not present

RFC-0008-plan-checker.md
```diff
@@ -0,0 +1,137 @@
...
+
+An SPI will be added that will add more checks to the `PlanChecker` class.  The SPI will contain a field indicating which phase
+of the plan checker to be added, and a validator that will be run during that phase.  The SPI will be added to the `PlanChecker`
+class, and the `PlanChecker` class will be modified to run the SPIs during the appropriate phase.
+
+A new endpoint to the Presto sidecar will be added that will attempt to convert a Presto plan to a Velox plan.  In the
+`presto-native-plugin` module, a new implementation of the SPI will be added which will add a check to the `PlanChecker` class
+which will call the sidecar to attempt to convert the plan.  If the conversion fails, the plan checker will fail the plan.  It will
+be added at the `validateFragmentedPlan` phase--this is because it's not until plan fragmentation occurs that we know which portions
+of the plan will be executed in the coordinator, and which will be executed in the workers.
```
BryanCutler, line 65: Plan fragments that are not meant for a worker are skipped, within the plugin

RFC-0008-plan-checker.md
```diff
@@ -0,0 +1,137 @@
...
+
+* `validateIntermediatePlan`
+* `validateFinalPlan`
+* `validateFragmentedPlan`
+
+It is generally useful to allow this class to be configured with an SPI that allows for custom plan validation logic to be added.
+For example, a business may decide that a certain type is not allowed, or add a check to ensure that plans that are overly
+complicated are killed.
+
+An SPI will be added that will add more checks to the `PlanChecker` class.  The SPI will contain a field indicating which phase
```
BryanCutler, line 57: Just to clarify the SPI will extend the built-in `PlanChecker` by allowing additional `Checker`s for each of the plan phases.

BryanCutler (commented):
Looks good, thanks!

RFC-0008-plan-checker.md
```diff
@@ -0,0 +1,137 @@
...
+> HTTP verb: POST
+>
+> Request body: serialized plan fragment
+>
+> Response body: JSON response indicating success, or if failure, the reason for the failure
+
+#### Additional information
+
+1. What modules are involved
+   2. `presto-native-sidecar`
```
rschlussel, line 83: this module doesn't exist yet?

RFC-0008-plan-checker.md
```diff
@@ -0,0 +1,137 @@
...
+* `validateIntermediatePlan`
+* `validateFinalPlan`
+* `validateFragmentedPlan`
+
+It is generally useful to allow this class to be configured with an SPI that allows for custom plan validation logic to be added.
+For example, a business may decide that a certain type is not allowed, or add a check to ensure that plans that are overly
+complicated are killed.
+
+An SPI will be added that will add more checks to the `PlanChecker` class.  The SPI will contain a field indicating which phase
+of the plan checker to be added, and a validator that will be run during that phase.  The SPI will be added to the `PlanChecker`
```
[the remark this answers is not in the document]
rschlussel: or rather, i think custom plan checkers are useful, but specifically velox plan translation could be independently useful and possibly faster to implement as an extension of explain.

RFC-0008-plan-checker.md
tdcmeehan: This will be a new module that is also being developed for other purposes such as function validation.  I will make that note in this RFC.

RFC-0008-plan-checker.md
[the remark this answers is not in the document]
rschlussel: I got this info from Sergey when I asked about it:

> Velox has "struct PlanFragment", which is what you are after. Adding toString() or toJson() method to it might work. PlanFragment has  std::shared_ptr<const core::PlanNode> planNode; which has rich toString() already.


I'm okay with both approaches, but let's make the post request be to a path specific for validation, so in the future we could extend to be able to get the full serialized plan if we want to.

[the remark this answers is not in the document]
tdcmeehan: I also see that PlanFragment is already serializable as a string.

I may not quite understand your original suggestion.  We are using a plan checker to provide validation of the plan prior to query execution, so we don't wait until execution to discover that the generated plan won't be able to execute.

This is tied to the lifecycle of the query.  It sounds like what you're suggesting is, add a new explain plan type that attempts to translate the plan to a Velox plan, and throw if it's not possible?  If so, I think we can do something along those lines too.

1) The new explain type could use the sidecar and SPI created in this proposal to accomplish this.
2) The output may contain a combination of Velox and Presto plans, however, since some fragments are coordinator only.

I think it's a good idea and we can include that in this proposal, as it would also allow for a non-intrusive way to determine if the plan could be executed in Presto C++.  However, I wanted to make sure this is what you're proposing.

RFC-0008-plan-checker.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### REST endpoint
+
+> Endpoint: /v1/plan
+>
+> HTTP verb: POST
+>
+> Request body: serialized plan fragment
+>
+> Response body: JSON response indicating success, or if failure, the reason for the failure
```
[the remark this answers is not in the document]
tdcmeehan: Agreed, will update.

RFC-0008-plan-checker.md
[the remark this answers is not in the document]
rschlussel: yeah, exactly.  The benefit is a bit tangential to the goal of query validation, as with the plan checker approach we can still validate the query by running explain(type distributed) with the plan checker enabled, but since we're doing all the work to translate the query plan, we may as well add the ability to print it.

RFC-0008-plan-checker.md
tdcmeehan: I've added this to the adoption plan.

tdcmeehan:
@rschlussel @BryanCutler I have updated the RFC with your feedback, thank you.

RFC-0008-plan-checker.md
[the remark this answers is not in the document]
tdcmeehan: I have added a section on the extended plan format.

RFC-0008-plan-checker.md
```diff
@@ -0,0 +1,171 @@
...
+be run before the query is queued.  This will allow the user to get feedback on the plan before the query is executed, and will
+allow the query to fail quickly if the plan is invalid.
+
+Because the queue limits concurrency, and too much concurrency during planning may require excessive resources in the coordinator,
+the plan checker will be run in a separate thread pool.  This thread pool will be configured with a maximum number of threads
+that can be run concurrently.  If the thread pool is full, then planning will wait until there is a free thread to run the planner.
+This will be configured with a new configuration parameter and session property.
+
+### EXPLAIN (TYPE DISTRIBUTED, FORMAT NATIVE)
+
```
rschlussel, line 96: format should specify how the data is represented (like text or json). Type should tell you what kind of data you'll get.  We should instead add a new type `EXPLAIN(TYPE NATIVE)` that will produce the native plan (which also happens to be distributed by definition). We might provide the option of outputting the Velox plan in text or JSON format, or even graphviz if someone feels like implementing it.

tdcmeehan: I thought about this, but decided against it:

1) Because the Velox plans are generated per fragment, it's really the fragmented plan that we are generating, which is what the distributed type gives you.
2) `FORMAT NATIVE` is meant to signify, just delegate the formatting to the underlying plugin, which in this case is just delegating to the plan converter.

Originally, I was unsure how to handle the `FORMAT` argument if specified.  But now that I think about it, perhaps what we can do is:

1) Add a new type, `DISTRIBUTED_NATIVE` or some such.  It's verbose, but it signifies that this is the distributed plan that will actually show the underlying eval's plan fragment, not Presto's version of it.
2) Add a format argument to the SPI so that we can configure the plan converter to generate different formats like Graphviz and JSON.

What do you think?

rschlussel: first want to note that we aren't just delegating the formatting to the underlying plugin. We are generating a new velox plan based on each plan fragment and returning the representation of that new plan.

For your suggestions:
For 1. I prefer  `NATIVE` to `DISTRIBUTED_NATIVE` as the new type because it's less verbose, and the native plan will always be distributed because it is based on plan fragments. However, either is fine.
For 2. sounds good to me

tdcmeehan: I updated the proposal to `EXPLAIN (TYPE NATIVE)`, and added details on how the format string would work, PTAL!

rschlussel (approved):

RFC-0008-plan-checker.md
```diff
@@ -0,0 +1,180 @@
...
+allow the query to fail quickly if the plan is invalid.
+
+Because the queue limits concurrency, and too much concurrency during planning may require excessive resources in the coordinator,
+the plan checker will be run in a separate thread pool.  This thread pool will be configured with a maximum number of threads
+that can be run concurrently.  If the thread pool is full, then planning will wait until there is a free thread to run the planner.
+This will be configured with a new configuration parameter and session property.
+
+### EXPLAIN (TYPE NATIVE)
+
+The `EXPLAIN (TYPE NATIVE)` command will be updated to run the `PlanConverter` over all fragments which are not `COORDINATOR_ONLY`.
```
BryanCutler, line 97: Should we make the endpoint take an array of fragments? We could just pass one at a time for the plan validation, but if EXPLAIN is able to gather all fragments, then it could be sent in a single request.

tdcmeehan: One by one keeps the API simple and given it's an explain plan, not really performance sensitive (and it's O(# of fragments), and realistically we'll never have > 100 fragments).

BryanCutler (approved):

PR #29, opened 2024-10-23
Add binary exchange RFC 0009

dan13bauer:


dan13bauer:
@aditi-pandit @majetideepak This is the RFC for the binary exchange protocol. Can you please take a look before I move it out of the draft state? Thank you!

tdcmeehan:
Does BinX provide security and does it encrypt the exchange?

dan13bauer:
> Does BinX provide security and does it encrypt the exchange?

It supports TLS encryption. Whenever SSL is configured for the HTTP server (http-server.https.enabled), it uses the same configuration and establishes TLS connections for the exchange.

tdcmeehan:
I understand why, in theory, using TCP sockets should be faster than HTTP.  But can you go into more depth on where we are seeing the speedup from: is it the lack of headers on the page data?  How the data is chunked in HTTP?

RFC-0009/RFC-0009-binary-exchange.md
```diff
@@ -0,0 +1,265 @@
...
+
+|Field name| Type | Semantic |
+| ---- | ---- | ---- |
+|replyType| enum (DATA, ACK, DELETE) | The type of reply |
+|status| enum (OK, NODATA, SERVER_ERROR, TIMEOUT) | The reply status |
+|bufferComplete| bool | True if the entire buffer has been transferred |
+|pageToken|uint64_t| The token (sequence number) of the first data page in the reply|
+|pageNextToken|uint64_t| The token (sequence number) of the next available page that can be requested|
+|remainingBytes| std::vector<uint64_t> | Contains the sizes of available pages when requested|
+|data| IOBuf | Contains the data payload |
```
tdcmeehan, line 125: Slightly ignorant question, but by encoding this field parallel to the other fields, does it require a memcpy?

dan13bauer:
> I understand why, in theory, using TCP sockets should be faster than HTTP. But can you go into more depth on where we are seeing the speedup from: is it the lack of headers on the page data? How the data is chunked in HTTP?

I don't have a good answer to that. The initial starting point was an observation that the CPU load of the HTTP stack is relatively high. @yingsu00, you have observed that recently, too, right? Our assumption was that a simpler stack should be able to provide some improvement but we didn't conduct a detail analysis yet.

RFC-0009/RFC-0009-binary-exchange.md
```diff
@@ -0,0 +1,265 @@
...
+It implements the exchange service using the BinX protocol. Incoming requests are forwarded to the
+`TaskManager` and the result is marshalled into a reply and sent back.
+
+The binary exchange server is started in `PrestoServer::run()`. It uses a dedicated
+IO thread pool in order to not interfere with the HTTP IO thread pool. The CPU thread pool is shared
+with the HTTP exchange.
+
+#### Implementation Notes
+
+The BinX server uses Wangle. It consists of the following components that are implemented in
```
majetideepak, line 146: Can you clarify how the Wangle stack is different from the Proxygen stack currently used by the HTTP implementation?
From a high-level, the asynchronous TCP socket handler `folly::AsyncServerSocket` is coming from folly in both stacks?

RFC-0009/RFC-0009-binary-exchange.md
```diff
@@ -0,0 +1,265 @@
...
+that gets called by the Velox exchange client and interfaces with Velox's page memory management. It also implements
+the exchange protocol logic.
+The `BinaryExchangeSource` uses the `BinaryExchangeClient` that is responsible for the protocol mechanics and
+implements connection setup, sending and receiving of requests and replies, and timeouts when replies don't arrive.
+
+The `PrestoServer` registers a factory method for creating exchange sources. This factory method has been extended
+such that `BinaryExchangeSource`s are created instead of HTTP exchanges when enabled by configuration.
+One exception are connections to the
+Presto coordinator that always uses the HTTP based exchange protocol. In a Kubernetes environment with its virtual
+networking, it is unfortunately not straight forward to detect whether the target host is the Presto connector
```
majetideepak, line 178: connector -> coordinator?

RFC-0009/RFC-0009-binary-exchange.md
```diff
@@ -0,0 +1,265 @@
...
+since the connector's service IP used in the Presto configuration doesn't correspond to the IP address used by the
```
majetideepak, line 179: connector -> coordinator?

majetideepak (commented):
Some nits and questions.
Like Tim, I am also curious to know where the wins are coming from given both Wangle and Proxygen use folly under the hood.

RFC-0009/RFC-0009-binary-exchange.md
```diff
@@ -0,0 +1,265 @@
...
+|pageNextToken|uint64_t| The token (sequence number) of the next available page that can be requested|
+|remainingBytes| std::vector<uint64_t> | Contains the sizes of available pages when requested|
+|data| IOBuf | Contains the data payload |
+
+`BinRequest` and `BinReply` are in-memory representations. The serialization and deserialization on top
+of TCP is implemented by the `ServerSerializeHandler` and `ClientSerializeHandler`. On the server side,
+requests are read and replies are written and vice-versa on the client side. During serialization, a
+length field is prepended that is needed to correctly de-serialize the protocol data units.
+
+The implementation uses Meta's [Wangle](https://github.com/facebook/wangle) framework.
```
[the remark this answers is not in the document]
dan13bauer: I believe Wangle is now an indirect dependency, both through Velox as well as through Proxygen. Do you think this needs to be made explicit?

RFC-0009/RFC-0009-binary-exchange.md
dan13bauer: Proxygen also uses Wangle underneath; at least portions of it. Wangle in turn uses folly's AsyncServerSocket. You are right, both stacks end up using folly's I/O mechanisms.
I'll add some clarifications.

RFC-0009/RFC-0009-binary-exchange.md
dan13bauer: Indeed! Thanks!

RFC-0009/RFC-0009-binary-exchange.md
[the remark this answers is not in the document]
majetideepak: You are right! I missed this.

PR #30, opened 2024-10-24
jdbc join push down

Ajas-MM:
IBM presto can push down the processing of  join queries, or part of  join queries into the connected jdbc data source. 

The results of this pushdown can include the following benefits:

- Improved overall query performance

- Reduced network traffic between IBM presto and the data source

- Reduced load on the remote data source 

- Significant cost reduction due to limited number of database hit


Ajas-MM:
This completed through another RFC https://github.com/Thanzeel-Hassan-IBM/rfcs/blob/main/RFC-0009-jdbc-join-push-down.md


https://github.com/prestodb/rfcs/pull/31

PR #35, opened 2025-01-24
JDK Upgrade RFC

aaneja:
RFC for JDK upgrade


RFC-0010-jdk-upgrade.md
```diff
@@ -0,0 +1,293 @@
...
+| 3.0.0 - 3.2.4 | Java 8/11 | 
+| 2.4.0 - 2.4.8 | Java 8 |
+
+
+This means, for example, if we upgrade the bytecode level of Presto to that of JDK 17, users would only be able to use Presto On Spark on Spark 3.3.0+ clusters
+
+### Workaround
+The [presto-spark](https://github.com/prestodb/presto/tree/498784bb35feaf2787da865865ffe8ab3e8673ca/presto-spark) maven module has the below dependency graph -
+
+```
```
ZacBlanco, line 69: For the rendered RFC, it might be better to put this in a ["details" block](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections)

RFC-0010-jdk-upgrade.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Proposed Migration Plan  
+
+To migrate to newer versions of the JDK, we propose a phased approach -   
+
+![slide from Prestocon 2024 presentation](RFC-0010/jdk_upgrade_timeline.png)
+
+The salient points - 
+
+ - During the upgrade, the project will be built with two JDK versions , the current and the new target version (e.g JDK 8 and 17) simultaneously
```
ZacBlanco, line 37: I think there could be cases in the future  in the upgrade where we may not be able to support both compilation and runtime of the two java versions. What happens in this case? Should we only support compiling with the older one, but attempt to support the newer runtime? 

I think we're lucky with the JDK 8/17 compatibility where we can do both, but in the case we can't we should have a fallback procedure - I believe we will hit this case when we go from Java 17 to 21 due to the behavior change in the decimal parsing. Running the same compiled code on different runtimes will give different results. I'm not sure if we want that.

RFC-0010-jdk-upgrade.md
```diff
@@ -0,0 +1,293 @@
...
+ - We will make use of our CI pipelines to run the full breadth of unit, integration and end-to-end product tests with every new JDK version
```
ZacBlanco, line 38: Maybe a little more specific with an example to make this clear?
```suggestion
 - We will make use of our CI pipelines to run the full breadth of unit, integration and end-to-end product tests on the newest JDK version that we intend to support. (e.g. in the case of 8 to 17 upgrade, we will run our CI on Java 17)
```

RFC-0010-jdk-upgrade.md
```diff
@@ -0,0 +1,293 @@
...
+ - Users will be invited to test new builds with their production workloads to report any issues observed around correctness, performance and stability
+ - As blocking issues shake out and get resolved, we will start deprecating support for older JDK versions
```
ZacBlanco, line 40: I think we should be specific about timelines (e.g. 1 release or 2 release version window where we attempt dual support before dropping support for the older JDK)

RFC-0010-jdk-upgrade.md
```diff
@@ -0,0 +1,293 @@
...
+Not using modern versions of our dependencies opens Presto installations to security risks through CVEs, which hurts adoption of Presto
+
+See this [Prestocon 2024 presentation](https://www.youtube.com/watch?v=K6CjuPZtqW4&list=PLJVeO1NMmyqW_qoMMEyVq-wW8lBSxA7hP&index=10) for a deep-dive motivating the issue further (Slides [here TODO]())
+
+
+## Proposed Migration Plan  
+
+To migrate to newer versions of the JDK, we propose a phased approach -   
+
+![slide from Prestocon 2024 presentation](RFC-0010/jdk_upgrade_timeline.png)
```
ZacBlanco, line 33: Since we're trying to be generic in this RFC I feel like it would be better to just include a table in text, and remove mention of the specific versions

| Phase | Time Window | Release Version | Runtime JDK Support | Compilation JDK Support |
|--------|--------|--------|--------|--------|
| Dual Support | Month N | v0.X | JDK Y and Y+1 | JDK Y |
| Dual Support & Deprecation | Month N + 2 | v0.(X+1) | JDK Y and Y+1 | JDK Y |
| Stable JDK | Month N + 4 | v0.(X+2) | JDK Y+1 | JDK Y+1 |

RFC-0010-jdk-upgrade.md
aaneja: Okay, added

RFC-0010-jdk-upgrade.md
aaneja: Cool, thanks for building the table!

RFC-0010-jdk-upgrade.md
aaneja: I think we will need to visit this case-by-case. For the decimal parsing one, we have to pick a behavior we want to support and call it out in our release notes. If we cannot force a behavior between two JDK versions, we document this caveat as well

RFC-0010-jdk-upgrade.md
```diff
@@ -0,0 +1,305 @@
...
+For the modules that *cannot be upgraded* to a higher bytecode level, we can refactor code to generate smaller modules that can be kept at lower bytecode levels.  
+This can prove to be beneficial for other modules too, e.g `presto-client` can be kept at JVM 8 bytecode level to allow for a broader support for Java based clients
+
+## Timeline for the project
+
+A detailed timeline is out-of-scope for this RFC.  
+We will track project milestones and timeline on https://github.com/orgs/prestodb/projects/31/
+
+We have the below high-level timeline targets -
+1. Migrate to JDK 17 by H1 2025, to JDK 21 by end of 2025
```
rschlussel, line 301: jdk 21 required by end of 2025 or just jdk 21 compatible?

RFC-0010-jdk-upgrade.md
```diff
@@ -0,0 +1,305 @@
...
+
+For the modules that *cannot be upgraded* to a higher bytecode level, we can refactor code to generate smaller modules that can be kept at lower bytecode levels.  
+This can prove to be beneficial for other modules too, e.g `presto-client` can be kept at JVM 8 bytecode level to allow for a broader support for Java based clients
+
+## Timeline for the project
+
+A detailed timeline is out-of-scope for this RFC.  
+We will track project milestones and timeline on https://github.com/orgs/prestodb/projects/31/
+
+We have the below high-level timeline targets -
```
rschlussel, line 300: Are there any timeline expectations for jdk 11 upgrade for presto-on-spark? we expect to be ready for this some time in q3.

ZacBlanco: I don't think we have any requirements for a JDK 11 upgrade on the presto-on-spark components on our side at the moment. However, I think we should pull up the minimum version as soon as those deployments are ready to do so. We should also make sure that we socialize that plan to the wider community if that is the case

RFC-0010-jdk-upgrade.md
aaneja: Let's be aggressive and target switchover by end of year 2025

RFC-0010-jdk-upgrade.md
rschlussel: oh wow, i thought this was a requirement.  This is very good to know.

RFC-0010-jdk-upgrade.md
rschlussel: I don't want to commit to this without doing some investigation first. We hope to run on java 21, but I don't know if we'll run into any issues with internal dependencies.  Also, I remember some discussion about correctness differences with decimals. can you remind me what the issue here was?

ZacBlanco: > Also, I remember some discussion about correctness differences with decimals. can you remind me what the issue here was?

Presto relies on JDK-internal floating point number parsing methods for decimal types and maintaining precision. Jackson also depends on the same when parsing floating point values within JSON. So it affects the JSON type as well. In JDK 19, a breaking change introduced the floating-point parsing code causes Presto to fail some unit tests because of floating point precision issues introduced by the JDK library's new implementation. I was able to create a partial fix for decimal types. However, the JSON parsing I don't have a solution for because we don't have an easy way to inject a new floating point value parsing method within Jackson.

The following commit has the relevant changes that fix the decimal parsing for JDK 21+: https://github.com/prestodb/presto/pull/23937/commits/0986e2205e904888b3fd179db61c7a936cbc7bf9

RFC-0010-jdk-upgrade.md
ZacBlanco: Just to clarify: We prototyped a change (https://github.com/prestodb/presto/pull/24395) which would split `presto-main` into two modules. core bits which PoS relies on, and then another module which is responsible for all the Presto server components that rely on Jetty. If we merge that PR the expectation is the the core bits can stay on Java 8 for a short amount of time while users (Meta, etc) take time to upgrade their Spark deployments. Everything else that isn't core would go to Java 17+ ASAP. If we're able to get that PR merged and upgrade all downstream dependencies to Java 17+ I think we will resolve the most critical CVEs on our end. However, we should strive to pull up the language level for the core and PoS bits sooner rather than later.

RFC-0010-jdk-upgrade.md
ScrapCodes: Some other opensource projects including Apache Spark has dealt with this situation by having two (or even three) separate source paths e.g. where a java source file does not compile with version 8 and needs a separate java-version-8 specific code would stay in `java8/`  source path, and same file for java-version-17` version would stay in `java17/`. Can we extend this to solve other situations e.g. dealing with difference of decimal parsing logic ?

So sources may be segregated as follows:

1. Java sources files that work with both java 8 and 17, would go in
`src/main/java`
2. Java source files that work with only java8. These will not exist in `src/main/java`, but only in
`src/main/java8`. For example, the code that is affected by difference in decimal parsing can stay separate and some
3. Java source files that work with only java17. These will not exist in `src/main/java`, but only in
`src/main/java17`. The files that exist in `src/main/java8` need to have a version for `src/main/java17` as well.

Next we have maven profiles, -Pjava8 or -Pjava17. During compilation we add source of java-version specific files to the compile path based on the maven profile selected.

This can be done for all three as well, for example, -Pjava11,-Pjava8 and -Pjava17 (or even 21).

Disadvantages of this approach:

1. Overhead of maintaining two source paths, for example, a enhancement to decimal parsing logic needs to be done for both java 8 and java17. We can hit challenges related to missing feature parity.
2. Need a CI checking/testing for both versions, before PR merge.

rschlussel: I think that's a good approach if it becomes necessary. FWIW I'm not too concerned about the decimal parsing difference. We already have differences with floating point -> string between Java and c++, and soon all the computation part will be in c++ anyway.

ZacBlanco (approved):

rschlussel (approved):

rschlussel:
@aaneja should we merge this?

aaneja:
@rschlussel Yes this can be merged. I don't have write access to this repo / merge rights. Can you merge it ?

PR #36, opened 2025-02-17
RFC for supporting mixed case identifiers 

agrawalreetika:
RFC for supporting mixed case identifiers


RFC-0010-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,201 @@
...
+## Backward Compatibility Considerations
+
+* Existing connectors that do not implement normalizeIdentifier will default to lowercase normalization.
+* Any connectors requiring case preservation can override the default behavior.
+* A configuration flag could be introduced to allow backward-compatible identifier handling at the catalog level.
+
+## Test Plan
+
+* Ensure that existing CI tests pass for connectors where no specific implementation is added.
+* Add support for mixed-case identifiers in at least one JDBC connector (e.g., MySQL, PostgreSQL) and create relevant unit tests.
```
ScrapCodes, line 178:
```suggestion
* Add unit tests for testing mixed-case identifiers support in a JDBC connector (e.g., MySQL, PostgreSQL).
```

RFC-0010-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,201 @@
...
+* Reetika Agrawal
+
+## Summary
+
+Improve Presto's identifier (schema, table & column names) handling to align with SQL standards, ensuring better interoperability with case-sensitive
+and case-normalizing databases while minimizing SPI-breaking changes.
+
+## Background
+
+Presto treats all identifiers as case-insensitive, normalizing them (typically to lowercase). This creates issues when
```
ScrapCodes, line 16:
```suggestion
Presto treats all identifiers as case-insensitive, normalizing them to lowercase. This creates issues when
```

RFC-0010-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,201 @@
...
+querying databases that are case-sensitive (e.g., MySQL, PostgreSQL) or case-normalizing to uppercase (e.g., Oracle,
+DB2). Without a standard approach, identifiers might not match the actual names in the underlying data sources, leading
+to unexpected query failures or incorrect results. Additionally, inconsistent handling of delimited and non-delimited
+identifiers across different connectors further complicates cross-engine compatibility.
```
ScrapCodes, line 20: Can you elaborate more on `inconsistent handling of delimited and non-delimited
identifiers`

RFC-0010-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,201 @@
...
+
+The goal here is to improve interoperability with storage engines by aligning identifier handling with SQL standards
+while ensuring a seamless user experience. Ideally, the change should be implemented in a way that minimizes
+backward-compatibility-breaking changes to the SPI, allowing connectors to adopt the new approach without significant
```
ScrapCodes, line 24:
```suggestion
breaking changes to the SPI, i.e. allowing connectors to adopt the new approach without significant
```

RFC-0010-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Background
+
+Presto treats all identifiers as case-insensitive, normalizing them (typically to lowercase). This creates issues when
+querying databases that are case-sensitive (e.g., MySQL, PostgreSQL) or case-normalizing to uppercase (e.g., Oracle,
+DB2). Without a standard approach, identifiers might not match the actual names in the underlying data sources, leading
+to unexpected query failures or incorrect results. Additionally, inconsistent handling of delimited and non-delimited
+identifiers across different connectors further complicates cross-engine compatibility.
+
+The goal here is to improve interoperability with storage engines by aligning identifier handling with SQL standards
```
ScrapCodes, line 22: Goals can be listed as bullet points:

* Improve interoperability with storage engines i.e standardizing identifier handling (SQL Standard ref: [ANSI?](https://www.ibm.com/docs/en/netezza?topic=md-handle-sql-identifiers#c_dbuser_handle_sql_identifiers__title__2)).
* Backwards compatibility with existing system. Minimal or no breaking changes to SPI.

*

ScrapCodes (changes requested):
Hi @agrawalreetika , thanks for the detailed write up. Got minor comments!

RFC-0010-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,201 @@
...
+  - Matching identifiers to locate entities such as catalogs, schemas, tables, views.
+  - Resolving column names based on table metadata provided by connectors.
+  - Passing identifiers to connectors when creating new entities.
+  - Processing and displaying entity names retrieved from connectors, including column resolution and metadata introspection commands like SHOW and DESCRIBE.
+
+## Proposed Implementation
+
+#### Core Changes
+
+* In the common code path, make changes to pass original identifier (Schema, table and column names)
```
ScrapCodes, line 54:
```suggestion
* In the presto-spi, make changes to pass original identifier (Schema, table and column names)
```

agrawalreetika:
Thanks for your review, @ScrapCodes . I've updated the document based on your comments. Please check it at your convenience.

RFC-0011-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,220 @@
...
+  - Non-delimited identifiers may be converted to lowercase or another case.
+
+3. Case-Insensitive (e.g., Hive)
+  - Delimited and non-delimited identifiers are treated the same.
+  - Identifiers may be automatically converted to a specific case.
+
+Presto's behavior includes:
+
+- Delimited identifiers ("Identifier") and non-delimited identifiers (identifier) are converted to lowercase by default
+  unless a connector enforces a specific behavior.
```
ZacBlanco, line 56: > unless a connector enforces a specific behavior.

Where do we enforce this currently?

RFC-0011-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,220 @@
...
+- Identifiers are normalized when:
+  - Resolving schemas, tables, columns, views.
+  - Retrieving metadata from connectors.
+  - Displaying entity names in metadata introspection commands like SHOW TABLES and DESCRIBE.
+
+Presto handles identifiers in several ways:
```
ZacBlanco, line 62:
```suggestion
Presto uses identifiers in several ways:
```

RFC-0011-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,220 @@
...
+    }
+```
+
+ConnectorMetadata.java
+```java
+
+    /**
+     * Normalize the provided SQL identifier according to connector-specific rules
+    */
+    default String normalizeIdentifier(ConnectorSession session, String identifier, boolean delimited)
```
ZacBlanco, line 105: Consider expanding the javadoc here to include the parameters and their descriptions (specifically I am concerned about documenting the meaning of `delimited`. I actually think we may want to rename this to "escaped" or maybe "quoted" as discussed in the conversation with Anant

RFC-0011-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,220 @@
...
+BaseJdbcClient.java
+```java
+    @Override
+    public String normalizeIdentifier(ConnectorSession session, String identifier, boolean delimited)
+    {
+        return identifier.toLowerCase(ENGLISH);
+    }
+```
+
+Example - Connector specific implementation -
```
ZacBlanco, line 137: Consider adding an example for the use of the `delimited` (escaped) argument

ZacBlanco (changes requested):
Thanks for this detailed proposal. I do have one question about performance. Is this something you have tested in any prototype? Will delegating out to the connectors for normalization impose any significant overhead during the analysis phase?

RFC-0011-mixed-case-identifier-support.md
agrawalreetika: Its going to be in ConnectorMetadata API implementation here - https://github.com/prestodb/rfcs/pull/36/files#diff-217aa5922002c9dabc8df724c3930f3fa969a98881495ad2648be620fdb50407R105

RFC-0011-mixed-case-identifier-support.md
agrawalreetika: Ok, we can do that. This is going to be a boolean for `double quotes`. And if in any connector we want to handle with and without quote differently https://github.com/prestodb/rfcs/pull/36/files#diff-217aa5922002c9dabc8df724c3930f3fa969a98881495ad2648be620fdb50407R30

I chose `delimited` currently since Identifier here defines it as `delimited` https://github.com/prestodb/presto/blob/master/presto-parser/src/main/java/com/facebook/presto/sql/tree/Identifier.java#L57 But I am open for suggestions for naming.

agrawalreetika:
Thanks for your review @ZacBlanco, I have addressed your comments. Please check.

Also about performance, I don't have the numbers yet but here its gonna add additional API call i.e. `normalizeIdentifier`

RFC-0011-mixed-case-identifier-support.md
ScrapCodes: Delimited is widely used in other databases too e.g. [Netezza](https://www.ibm.com/docs/en/iis/11.7?topic=connector-netezza-identifiers)
But, Agree with Zac about documenting the meaning of delimited, and how it is a better alternative to escaped. Where each special character needs to be escaped and in case of delimited only the delimiter needs to be escaped.

RFC-0011-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,234 @@
...
+
+Presto treats all identifiers as case-insensitive, normalizing them to lowercase. This creates issues when
+querying databases that are case-sensitive (e.g., MySQL, PostgreSQL) or case-normalizing to uppercase (e.g., Oracle,
+DB2). Without a standard approach, identifiers might not match the actual names in the underlying data sources, leading
+to unexpected query failures or incorrect results. Additionally, inconsistent handling of delimited and non-delimited
+identifiers across different connectors further complicates cross-engine compatibility.
+
+The goal here is to improve interoperability with storage engines by aligning identifier handling with SQL standards
+while ensuring a seamless user experience. Ideally, the change should be implemented in a way that minimizes
+breaking changes to the SPI, i.e. allowing connectors to adopt the new approach without significant.
```
ScrapCodes, line 24:
```suggestion
breaking changes to the SPI, i.e. allowing connectors to adopt the new approach without significant impact.
```

RFC-0011-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,234 @@
...
+ sys                
+ testdb             
+(6 rows)
+
+presto> show tables from mysql.TestDb;
+   Table   
+-----------
+ Test      
+ TestTable 
+ testtable
```
ScrapCodes, line 169: Sounds like, we are making presto case sensitive while remaining backwards compatible. To me the term "Mixed case" across the doc sounds less specific, not sure what others think, but `case sensitive` is more widely used and precise.

ScrapCodes (commented):
Thanks for quick turn around!, 

Do you think we can call this RFC: "Standardize the handling of delimited identifiers" or "Support case sensitive identifiers in Presto". To me case sensitive is a better term than Mixed case.
Also adding clarification about what delimited identifier means, and how it is a better option over escaped identifiers. Where delimited identifier needs only delimiter to be escaped, whereas an escaped identifier needs every special character escaped.

RFC-0011-mixed-case-identifier-support.md
agrawalreetika: I think we shoun't call it escaped since this is going to be only for `double quotes` boolean. May be we can rename boolean to `quoted` if that makes more clear? Open to hear suggestions here?

agrawalreetika:
Thank for your suggestion @ScrapCodes 
If it makes it more clear, than we can refactor RFC header from `Mixed case identifiers` to  `Support case-sensitive identifiers in Presto`?

RFC-0011-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,234 @@
...
+To ensure backward-compatibility current connectors where connector specific implementation is not added, existing CI tests should pass.
+Add support for mixed case for a JDBC connector (ex. mysql, postgresql etc) and add relevant Unit tests for same.
+
+## Modules involved
+- `presto-main`
+- `presto-common`
+- `presto-spi`
+- `presto-parser`
+- `presto-base-jdbc`
+
```
ScrapCodes, line 227: Consider adding the link to the POC implementation, if you have one.

RFC-0011-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,205 @@
...
+* A configuration flag could be introduced to allow backward-compatible identifier handling at the catalog level.
+
+## Test Plan
+
+* Ensure that existing CI tests pass for connectors where no specific implementation is added.
+* Add unit tests for testing mixed-case identifiers support in a JDBC connector (e.g., MySQL, PostgreSQL).
+* Cover cases such as:
+  - Queries with mixed-case identifiers.
+  - Metadata retrieval commands (SHOW SCHEMAS, SHOW TABLES, DESCRIBE).
+  - Joins, subqueries, and alias usage with mixed-case identifiers.
```
aaneja, line 184: Can also add `Queries between two connectors with different normalization rules, e.g MySQL and Hive`

RFC-0011-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,205 @@
...
+  case-normalizing databases.
+- Minimize SPI-breaking changes to maintain backward compatibility for existing connectors.
+- Introduce a mechanism for connectors to define their own identifier normalization behavior.
+- Allow identifiers to retain their original case where necessary, preventing unexpected query failures.
+- Ensure Access Control SPI can correctly normalize identifiers.
+- Preserve a seamless user experience while making these changes.
+
+### Proposed Plan
+
+Presto's behavior includes:
```
aaneja, line 37: nit: 
```suggestion
Presto's default behavior is -
```

ScrapCodes (approved):

RFC-0011-mixed-case-identifier-support.md
```diff
@@ -0,0 +1,205 @@
...
+
+### Goals
+
+- Align Presto’s identifier handling with SQL standards to improve interoperability with case-sensitive and
+  case-normalizing databases.
+- Minimize SPI-breaking changes to maintain backward compatibility for existing connectors.
+- Introduce a mechanism for connectors to define their own identifier normalization behavior.
+- Allow identifiers to retain their original case where necessary, preventing unexpected query failures.
+- Ensure Access Control SPI can correctly normalize identifiers.
+- Preserve a seamless user experience while making these changes.
```
hantangwangd, line 33: @agrawalreetika thanks for this very helpful proposal. Bring a few questions for discussing. Please let me know if there is anything I didn't understand correctly.

It seems that when setting `case-sensitive-name-matching=true`, the effect of delimited identifier on `catalog/schema/table` is the same as that of undelimited identifier. This means that unlimited identifiers also retain the case, which seems a little different from the SQL spec. So should we add some explanations about the behavior of delimited and undelimited identifiers?

Besides, for databases like PostgreSQL, it's case sensitive for delimited column names, which means that the table can contain both column "ABCol" and column "AbCol" at the same time. Have we considered this situation? If not, should we clearly state that this situation is not supported in presto?

hantangwangd: @agrawalreetika I saw the PR first, and then came to read the proposal :smile:. The column issue is OK as long as we're explicitly aware of the current limitations.

I'm a little concerned about how we support `Postgresql`. Basically, when it comes to case sensitivity, the default behavior of `Postgresql` follows the SQL specification most faithfully, that is, identifiers in quotation marks are treated as case-sensitive in SQL statement, whereas unquoted identifiers are automatically converted to lowercase.

So, do you think it makes sense to pass the state of `delimited` as a parameter when define the SPI methods? For example:

Metadata.java

```java
    String normalizeIdentifier(Session session, String catalogName, String identifier, boolean delimited);
```

ConnectorMetadata.java
```java

    /**
     * Normalize the provided SQL identifier according to connector-specific rules
    */
    default String normalizeIdentifier(ConnectorSession session, String identifier, boolean delimited)
    {
        return identifier.toLowerCase(ENGLISH);
    }
```

As I understand, if some connectors (such as Postgresql) is unaware of this `delimited` flag, it cannot make the right case-handling logic. Other connectors (such as Mysql) may optionally ignore this flag.

Haven't dug deep into the code, so not very sure about the implementation complexity. Please let me know if this is a feasible way.

agrawalreetika: Thanks again for the insightful follow-up, @hantangwangd!

You're right about the importance of the delimited flag, especially for connectors like PostgreSQL that handle identifier casing differently based on whether the identifier is quoted.

In fact, during the initial proposal, I had a similar thought and included the delimited flag as a parameter in the proposed SPI method for exactly this reason—to give connectors enough context to apply the correct case-handling logic based on whether the identifier was quoted or not.

However, in the current state of Presto, the information about whether a schema or table name was quoted in the original SQL isn't preserved by the time it reaches the SPI layer. Adding support for retaining that quoting information would require broader changes across the planning and analysis layers. So I scoped that out for now and focused on getting the base support in place, with the intention to handle quoting and delimited identifiers more fully in a follow-up iteration.
You're absolutely right that for PostgreSQL and similar connectors, this context is important.

Let me know if that sounds reasonable, or if you'd prefer we try to tackle quoting preservation sooner.

hantangwangd: @agrawalreetika Thanks for your explanation.

As you said, the delimited flag is very important for handling the case sensitivity of identifier, so IMO we should at least take it into account in the proposal, since the solution designed in the proposal should be as comprehensive as possible.

For actual implementation, I agree with you that we can accomplish the whole proposal progressively though several PRs ---- especially if retaining the quoting information for schema/table name need a lot of changes. Do you think this makes sense?

aaneja: >I'm a little concerned about how we support Postgresql. Basically, when it comes to case sensitivity, the default behavior of Postgresql follows the SQL specification most faithfully, that is, identifiers in quotation marks are treated as case-sensitive in SQL statement, whereas unquoted identifiers are automatically converted to lowercase.

@hantangwangd Since we use quoting for [any SQL we generate from the BaseJdbcClient](https://github.com/prestodb/presto/blob/39594cc518f0bea74e3881d6cfd05988cf85e4f0/presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java#L299-L303) any identifier passed (for schema or catalog name) will always be quoted in the SQL submitted to Postgres.

With the existing interface, if the Postgres connector chooses to not normalize its identifiers to lower-case we would get the below behavior (once we fix up the normalization for column names too) - 
For a DDL submitted to Presto -
```
create table test1
(
UPR integer,
lwr integer,
Mixed integer,
"Select" integer
);
```
the underlying DDL submitted to Postgres would be -
```
create table "test1"
(
"UPR" integer,
"lwr" integer,
"Mixed" integer,
"Select" integer
);
```

To me, this seems like reasonable and good-to-have behavior - users get exactly the case that they submitted for the idnetifiers

If the user does want lower-case column names in Postgres for the table created they do have the option of making this explicit in the DDL they submit to Presto -
```
create table test1
(
upr integer,
lwr integer,
mixed integer
);
```

hantangwangd: Hi @aaneja , thanks for sharing your perspective and the example, so we can discuss it a little more based on this case. If I were a user and believed that Presto is `align with SQL standards`, then when I execute the following DDL:

```
create table test1
(
UPR integer,
lwr integer,
Mixed integer,
"Select" integer
);
```

The statement I actually want to send to PostgreSQL was:

```
create table "test1"
(
"upr" integer,
"lwr" integer,
"mixed" integer,
"Select" integer
);
```

Because the SQL standard specifies that unquoted identifiers are case-insensitive in comparisons, and PostgreSQL uniformly converts all unquoted identifiers to lowercase.

I agree that the behavior you described above is reasonable and do not have ambiguity, but it doesn't fully align with the SQL standard. So do we want to go with it? If we do, we might need to document it clearly in some place of the user docs.

What do you think? Please let me know if there's anything I misunderstood.

hantangwangd: > We could call this out in the RFC and point to these examples as what behavior to expect

@aaneja 100% agree with this, it's always helpful to demonstrate the specific expected behavior through concrete examples.

> Do you see any gotchas for SELECT / DML statements though ?
> I think this RFC is primarliy enabling reading case-sensitive schemas/tables/columns and we should shake out those issues first

Seems that the core of our discussion revolves around how we intend to handle unquoted identifiers. If the goal is to quickly enable reading case-sensitive schemas/tables/columns, I agree that — with `case-sensitive-name-matching` flag be enabled — treating unquoted identifiers as strictly case-sensitive is a reasonable way, and the approach is OK for now since the parsing and maintaining of the delimited flag for schema/table identifiers would likely require a lot of efforts.

In future, if we implement maintaining and handling of the delimited flag for all identifiers, it seems that we no longer need this dedicated `case-sensitive-name-matching` configuration. At that point, it could be more align with SQL standard behavior: when users use unquoted identifiers, it implies case insensitivity (with specific case conversion handled by each connector); whereas when users use quoted identifiers, they meant to enforce strict case sensitivity. FYI, this behavior has been discussed for the case sensitivity of `RowType` field names and `JSON` column names, see [here](https://github.com/prestodb/presto/pull/21869#issuecomment-1933265932) and [here](https://github.com/prestodb/presto/pull/21869#issuecomment-1936579389).

agrawalreetika: Thanks for your pointers and review @aaneja & @hantangwangd. I will add the example as well in the RFC for clarity. 

> In future, if we implement maintaining and handling of the delimited flag for all identifiers, it seems that we no longer need this dedicated case-sensitive-name-matching configuration. 

My understanding here is, once we start sending `delimiter` flag as well for `normalizeIdentifier` API then in case of `delimiter = false` it could be legacy (converting it to lower case) and `delimiter = true` it would be converted to connector specific case. We don't have to have `case-sensitive-name-matching` configuration check in the `normalizeIdentifier` API.

Current behaviour is -
If `case-sensitive-name-matching = false`
- converts schema/table names to lower case with or w/o delimiter (Existing behaviour)

`case-sensitive-name-matching = true `
- converts schema/table names to connector-specific casing with or w/o delimiter

hantangwangd: @agrawalreetika, thanks for your summarizing, great understanding overall, just on little point to clarify when we start sending delimiter flag as well for `normalizeIdentifier` API.

As I understand, we send the original-case identifier along with its delimiter flag to the `normalizeIdentifier` API, and then in case of,

 - `delimiter = false`: the identifiers need to be normalized. Presto's default behavior is lowercase conversion, but connectors can customize their own normalization logic ---- like Oracle uppercasing them, PostgreSQL lowercasing them, and MySQL keep the case as-is.
 - `delimiter = true`: the identifiers always need to preserve their original case.

Do you think this behavior makes sense? If there's anything I misunderstood, please let me know.

agrawalreetika: Sounds fair to me. 
We will also have flexibility in here, since we have `normalizeIdentifier` API layout already. In case we still want to have a flag to keep the current legacy behavior (lower-case), we would have that option as well.

@aaneja @hantangwangd I’ve added detailed Behavioral Examples to the RFC based on our earlier discussion. Please take a look at your convenience and share any feedback.

Also, when you get a chance, kindly drop a review on the corresponding PR: https://github.com/prestodb/presto/pull/24551. Thanks a lot!

hantangwangd (approved):
Thanks for clarifying the behaviors of both quoted and unquoted identifiers with `case-sensitive-name-matching` flag.

agrawalreetika:
@tdcmeehan, this requires review and merge from someone with write access.

PR #37, opened 2025-04-03
Add the CLP connector RFC

wraymo:
This PR adds an RFC for the CLP connector, which enables SQL queries on the compressed CLP archives.

### DIFF
--- a/RFC-0012-clp-connector.md
+++ b/RFC-0012-clp-connector.md
@@ -1,0 +1,267 @@
+# **RFC0012 for Presto**
+
+## CLP connector
+
+Proposers
+
+* Rui Wang ([@wraymo](https://github.com/wraymo)), YScope
+* Devin Gibson ([@gibber9809](https://github.com/gibber9809)), YScope
+* Xiaochong Wei ([@anlowee](https://github.com/anlowee)), YScope
+* Yu(Jack) Luo ([@jackluo923](https://github.com/jackluo923)), YScope
+* Kirk Rodrigues ([@kirkrodrigues](https://github.com/kirkrodrigues)), YScope
+
+## Related Issues
+
+* [y-scope/clp#78](https://github.com/y-scope/clp/issues/78)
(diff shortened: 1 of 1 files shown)


RFC-0012-clp-connector.md
```diff
@@ -0,0 +1,267 @@
...
+| `CLP_GET_JSON_STRING()`             | Converts an entire log record into a JSON string.                                                      | `VARCHAR`              |
+| `CLP_GET_STRING_ARRAY(<json-path>)` | Retrieves an array value corresponding to the given JSON path and converts each element into a string. | `ARRAY(VARCHAR)`       |
+| `CLP_WILDCARD_COLUMN()`             | Used in filter expressions to allow comparisons with *any* column in the log record.                   | N/A                    |
+
+**Table 3:** Useful CLP UDFs for working with schemaless data in CLP.
+
+## Table & Column Resolution
+
+Implementing `ConnectorMetadata` requires determining what tables and columns exist in a CLP dataset, which in turn requires querying CLP's metadata database. Since different users may have their own implementation of a CLP metadata database, the connector should expose the required methods through a `ClpMetadataProvider` interface. For table and column resolution, the interface should contain the following methods:
```
tdcmeehan, line 105: To be clear, `ClpMetadataProvider` lives in the CLP connector, not the runtime?

tdcmeehan (approved):
This all seems pretty reasonable to me.  There doesn't seem to be anything that needs to change in the SPI, runtime or execution engines besides things that are at the connector level.

One suggestion is to keep the CLP C++ connector in Presto for now.  We can always move it to Velox later, and it will make the landing of this code more predictable.

wraymo:
> This all seems pretty reasonable to me. There doesn't seem to be anything that needs to change in the SPI, runtime or execution engines besides things that are at the connector level.
> 
> One suggestion is to keep the CLP C++ connector in Presto for now. We can always move it to Velox later, and it will make the landing of this code more predictable.

Thanks for the suggestion! We'll spend some time working on it.

PR #39, opened 2025-06-10
Add Presto TVF RFC

aditi-pandit:


jaystarshot:
LGTM, it would be great if we could split the RFC into what is configurable and what is built in so that its easier for the users to read the RFC directly

RFC-0011-tvf.md
```diff
@@ -0,0 +1,737 @@
...
+Table functions are implemented through connectors. A TableFunction derives from AbstractConnectorTableFunction.
+
+```java
+public abstract class AbstractConnectorTableFunction
+       implements ConnectorTableFunction
+{
+   private final String schema;
+   private final String name;
+   private final List<ArgumentSpecification> arguments;
+   private final ReturnTypeSpecification returnTypeSpecification;
```
jaystarshot, line 164: What is the `ReturnTypeSpecification` ?

aditi-pandit:
> LGTM, it would be great if we could split the RFC into what is configurable and what needs to be configured and what is built in so that its easier for the users to read the RFC directly

I had split the declaration, analysis parts of the API and the execution API with the planning pieces (from the framework) in between. I can see why this is confusing as the user does not have continuity with the API. Have changed the writing now.

tdcmeehan (commented):
Can you add more details from the template:

1) Adoption plan (I expect this will be simple, opt-in for all connectors and new TVFs)
2) Test plan
3) Other approaches considered

RFC-0011-tvf.md
```diff
@@ -0,0 +1,752 @@
...
+
+All columns of the table are pass-through if this property is set. Else only partition columns of the table are pass-through.
+
+```java
+public class TableArgumentSpecification
+       extends ArgumentSpecification
+{
+   private final boolean rowSemantics;
+   private final boolean pruneWhenEmpty; 
+   private final boolean passThroughColumns;
```
tdcmeehan, line 372: While I can't think of a way to simplify this, this seems complicated to debug if the connector author doesn't get it right.  I'm wondering what sort of testing tools we can give to connector authors to validate the behavior of their functions?  For example, is it possible to see the plan of a proposed table valued function and validate that the plan is correct?  It would seem like any TVF would want to write tests that validate the TVF is correctly planned, right?

aditi-pandit: Agree. It might be complicated for function authors to test different combinations of SQL to ensure they are planned and evaluated correctly. We might need to develop a fuzzer of sorts for it. I'll put some thought into how we can do it in a principled way.

PR #40, opened 2025-07-30
Add RFC for new SPI for SQL invoked functions

pdabre12:


RFC-0011-plugin-based-sql-invoked-functions-spi.md
```diff
@@ -0,0 +1,250 @@
...
+Previously, inlined SQL invoked functions were registered under the `presto.default` namespace (the default Java built-in namespace). Native functions pulled from the c++ sidecar were registered under a separate namespace like `native.default`.
+These native functions contain overridden optimized implementations of some of the inlined SQL invoked functions that the users would like to take advantage of. 
+However, this setup led to conflict resolution complexity. Presto would need to prioritize between two namespaces `presto.default` and `native.default` when resolving functions and such namespace-based prioritization is opaque and not ideal.
+
+To address this, the proposed change moves all inlined SQL invoked functions into a plugin-based namespace manager `BuiltInPluginFunctionNamespaceManager` which serves whatever the current default function namespace is. 
+This ensures:
+- All inlined SQL invoked functions are grouped together cleanly and loaded via a plugin.
+- Overridden native implementations of inlined SQL invoked functions are excluded from plugin registration.
```
feilong-liu, line 35: Will SQL invoked functions having native implementations be skipped? I am asking because I see in the current PR https://github.com/prestodb/presto/pull/25597/files it will throw error for duplicates.
I am wondering if we can skip if getting name conflicts, as it will not need manually removal of existing functions.

RFC-0011-plugin-based-sql-invoked-functions-spi.md
```diff
@@ -0,0 +1,250 @@
...
+### Goals
+- Introduce a new SPI method, `getSqlInvokedFunctionsSPI` for plugin-based registration of SQL invoked functions.
+- Create and wire a new plugin-based namespace manager `BuiltInPluginFunctionNamespaceManager` to manage these functions.
+- Enable native sidecar environments to load only the relevant subset of inlined SQL-invoked functions (i.e., excluding ones overridden natively).
+- Guarantee that inlined SQL invoked functions still work under the default catalog using the new plugin namespace manager.
+
+### Non-goals
+- This RFC does not Change the execution model for inlined SQL invoked functions - they are still evaluated in Java as earlier.
+- It does not modify function resolution rules outside of registration-time separation (i.e., no new prioritization logic is added).
+- It does not impact native (functions pulled from sidecar) or Java built-in function registration, which remains handled by the `NativeFunctionNamespaceManager` and `BuiltInTypeAndFunctionNamespaceManager` respectively.
```
feilong-liu, line 47: @kevintang2022  has a proposal which use this design to support native functions in Prestissimo https://docs.google.com/document/d/13tv1NYKWfy3v54kV1fyXFOYMHpCUkcEd6w8XSTyd2d0/edit?usp=sharing
which needs some extensions for this design.
I wonder if we can have one more built-in function plugin for native functions, so that native functions will also be under presto.default namespace, which will lead to seamless support for evaluation of functions of Java/Cpp/SQL at the same time.

feilong-liu (commented):
Looks good to me. Add one more comment on extensions of the design to better support Java/Cpp/SQL functions.

RFC-0011-plugin-based-sql-invoked-functions-spi.md
pdabre12: Yes, SQL invoked functions having native implementations will be skipped provided we load the `presto-native-sql-invoked-functions-plugin`, this is only supported for sidecar-enabled clusters.

> I am wondering if we can skip if getting name conflicts, as it will not need manually removal of existing functions.

Yes, this could be done but it makes sense only in the case of overridden native implementations because we prioritize those over the sql invoked fns. In other cases/ plugins loaded , there isn't a priority assigned to the Fns loaded via the plugin so I thought it's better to fail at duplicate signatures instead of silently skipping them. What do you think?

Also the functions loaded via the plugin will be available before the functions from the sidecar(as its a lazy fetch call), so you will need to remove that overridden plugin loaded sql invoked fn from `BuiltInPluginFunctionNamespaceManager` which isn't ideal.

feilong-liu (approved):

RFC-0011-plugin-based-sql-invoked-functions-spi.md
pdabre12: So we check for conflict mismatches only against the plugins and the default serving namespace manager. The code block where we do the conflict check.

```
private synchronized FunctionMap checkForNamingConflicts()
    {
        Optional<FunctionNamespaceManager<?>> functionNamespaceManager =
                functionAndTypeManager.getServingFunctionNamespaceManager(functionAndTypeManager.getDefaultNamespace());
        checkArgument(functionNamespaceManager.isPresent(), "Cannot find function namespace for catalog '%s'", functionAndTypeManager.getDefaultNamespace().getCatalogName());
        checkForNamingConflicts(functionNamespaceManager.get().listFunctions(Optional.empty(), Optional.empty()));
        return functions;
    }

    private synchronized void checkForNamingConflicts(Collection<? extends SqlFunction> functions)
    {
        for (SqlFunction function : functions) {
            for (SqlFunction existingFunction : this.functions.list()) {
                checkArgument(!function.getSignature().equals(existingFunction.getSignature()), "Function already registered: %s", function.getSignature());
            }
        }
    }
```
We would need to add code here to check signature conflicts for functions residing in the `BuiltInNativeFunctionNamespaceManager`.

 To clarify,  the checks are:
 1. Plugins with other plugins
 2. Plugins with `BuiltInNativeFunctionNamespaceManager`
 3. Plugins with current serving namespace manager
 
Note that we aren't checking for conflicts with `BuiltInNativeFunctionNamespaceManager` against the current serving namespace manager as in your case, that would lead to manual removal of functions, but we still wanna enforce the duplicate signatures not allowed in other places.

tdcmeehan (approved):

pdabre12:
@rschlussel Do you have any comments on this?

PR #41, opened 2025-08-05
Add rfc for native built in functions

kevintang2022:


kevintang2022:
> @kevintang2022 Just a very basic question, so with this change all the fns from the sidecar would be under `presto.default` namespace? Where does this new `BuiltInNativeFunctionNamespaceManager` fit with respect to the `NativeFunctionNamespaceManager`? (the function namespace manager which currently holds the sidecar functions)

Adding this should have no effect on the NativeFunctionNamespaceManager. NativeFunctionNamespaceManager is still the serving namespace manager for `native.default.`

The reason it is needed is because we want users to be able to call native functions without specifying the full qualified name. It seems in the current implementation of sidecar, users need to fully type `native.default.function_name` to invoke a C++ function if they are still using `presto.default.` as default namespace

pdabre12:
Thanks, I see.
However, switching of the namespaces from `presto.default` to `native.default` is something that is needed for sidecar-enabled native clusters. Why would we still be using `presto.default` namespace in a sidecar enabled native cluster?

kevintang2022:
> Thanks, I see. However, switching of the namespaces from `presto.default` to `native.default` is something that is needed for sidecar-enabled native clusters. Why would we still be using `presto.default` namespace in a sidecar enabled native cluster?

Actually, it's not required to use native.default with sidecar enabled. You can have presto.default and have sidecar enabled, and you can call the native functions by using fully qualified name. 

The reason this is needed is because constant folding is not quite ready to be used with the sidecar functions. This means that we still need to use coordinator implementation in some cases, but use worker implementation in some other cases.

You're right that if we just want to always use the worker implementation, then NativeFunctionNamespaceManager will work as is. However, we need to have some selective logic for the use case of sometimes using Java implementation and sometimes using C++ implementation

RFC-0012-built-in-native-function-namespace-manager.md
```diff
@@ -0,0 +1,101 @@
...
+* All functions with Java implementations get constant folded on the coordinator
+* All functions without Java implementations skip constant folding
+* Both functions with and without Java implementations get registered as "builtin", meaning that they don't need to be fully qualified to be used
+
+Breakdown by case
+* Function with only java implementation: constant fold, existing behavior
+* Function with only C++ implementation: no constant fold
+* Function with java and c++ implementation: constant fold, existing behavior
+* Function with only SQL implementation: use sql implementation, existing behavior
+* Function with SQL and C++ implementation: use c++ implementation, new behavior
```
pdabre12, line 36: > Function with SQL and C++ implementation: use c++ implementation, new behavior.

Do we have an example of this implementation?

kevintang2022:
> Have you tested the scenario where this exact setup works? Prestissimo doesn't support multiple namespaces. The functions that we report from Prestissimo need to be under a namespace, in the case where sidecar is enabled it is `native.default`. This namespace is configurable by the worker configuration property `presto.default-namespace`. Note that this property exists on the coordinator as well, which we won't be setting (default is `presto.default`).

Yes, I tested this with presto.default namespace as default config for both coordinator and worker. In this scenario, it's not possible to run native.defauolt functions, but everything else will be the same as if the sidecar was not enabled. This is because for most cases, we just want existing behavior. It's only for very few cases where we would want to change existing behavior (like C++ implementation overriding a SQL implementation)


> Additionally another question is, when you run queries let's say `abs()` without the fully qualified namespace, it will resolve it to the `presto.default` namespace and ideally we think this should be constant folded, but what if it wasn't constant folded? If it wasn't constant folded , we would send over that function with name as `presto.default.abs` over to the C++ worker side. Now, we set the worker configuration property `presto.default-namespace` to `native.default` and hence we would run into a function not registered error.

No, this would not throw error because we never use native.default. It will be presto.default for both coordinator and worker.

RFC-0012-built-in-native-function-namespace-manager.md
kevintang2022: map_remove_null_values has a SQL implementation in coordinator and C++ implementation from worker

pdabre12:
> Yes, I tested this with presto.default namespace as default config for both coordinator and worker. In this scenario, it's not possible to run native.defauolt functions, but everything else will be the same as if the sidecar was not enabled. This is because for most cases, we just want existing behavior. It's only for very few cases where we would want to change existing behavior (like C++ implementation overriding a SQL implementation)

I see this is for a specific use case, where in you don't want to use the `native.default` functions at all, but still pull in functions from the sidecar under the `presto.default` namespace. What this will enable you to do is use the native overridden implementations for the sql invoked fns and also give you access to the underlying native functions which don't have implementations in Java. Is my understanding correct?

RFC-0012-built-in-native-function-namespace-manager.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+## Proposed Implementation
+
+The approach will be somewhat similar to this PR: https://github.com/prestodb/presto/pull/25597
+
+Instead of adding a BuiltInPluginFunctionNamespaceManager, add a BuiltInNativeFunctionNamespaceManager. Register the functions from native sidecar under the presto.default namespace. During function evaluation inside of FunctionAndTypeManager#resolveFunctionInternal and getMatchingFunctionHandle, add some logic to determine whether the function in BuiltInTypeAndFunctionNamespaceManager or the function in BuiltInNativeFunctionNamespaceManager is used. This is similar to the PR, except the logic in PR will throw an error if it sees that there is a conflict. If there is a conflict, follow this logic to determine which functions to use:
+
+
+* If the one already inside the BuiltInTypeAndFunctionNamespaceManager is a Java implementation, then use the one in BuiltInTypeAndFunctionNamespaceManager
```
pdabre12, line 59: I guess you don't have access to the `FunctionMetadata` at this point to determine its implementation.

kevintang2022:
> > Yes, I tested this with presto.default namespace as default config for both coordinator and worker. In this scenario, it's not possible to run native.defauolt functions, but everything else will be the same as if the sidecar was not enabled. This is because for most cases, we just want existing behavior. It's only for very few cases where we would want to change existing behavior (like C++ implementation overriding a SQL implementation)
> 
> I see this is for a specific use case, where in you don't want to use the `native.default` functions at all, but still pull in functions from the sidecar under the `presto.default` namespace. What this will enable you to do is use the native overridden implementations for the sql invoked fns and also give you access to the underlying native functions which don't have implementations in Java. Is my understanding correct?

Yes that describes our use case

pdabre12 (approved):

rschlussel (approved):

tdcmeehan (changes requested):
This RFC doesn't have an adoption plan, [which is a part of the template](https://github.com/prestodb/rfcs/blob/main/RFC-0000-template.md#adoption-plan).  Can you add details on the eventual deprecation and removal of it (once constant folding fully delegates to C++ evaluation)?

aditi-pandit (commented):
Thanks for this writeup.

RFC-0012-built-in-native-function-namespace-manager.md
```diff
@@ -0,0 +1,108 @@
...
+
+* All functions with Java implementations get constant folded on the coordinator
+* All functions without Java implementations skip constant folding
+* Both functions with and without Java implementations get registered as "builtin", meaning that they don't need to be fully qualified to be used
+
+Breakdown by case
+* Function with only java implementation: constant fold, existing behavior
+* Function with only C++ implementation: no constant fold
+* Function with java and c++ implementation: constant fold, existing behavior
+* Function with only SQL implementation: use sql implementation, existing behavior
```
aditi-pandit, line 35: I would presume SQL inlining happens before constant folding. They are orthogonal concerns as far as I understand. SQL inlining issues are handled by https://github.com/prestodb/rfcs/pull/40/files only.
Can you confirm ? 

But there is a correlation in that :
- If there is a C++ function for a SQL function, then we want to pick up the C++ function. So if the function was originally constant folded it will not anymore. 
- If the SQL function expands to only java functions then it will get constant folded. 
- For new SQL functions that expand to new C++ functions, these are not constant folded.

That's about the impact to the current RFC.

RFC-0012-built-in-native-function-namespace-manager.md
```diff
@@ -0,0 +1,108 @@
...
+
+[Question]
+* Use presto.default namespace for every built in function or have separate namespace for native sidecar functions?
+
+[Answer]
+* If we choose to use presto.default. prefix for every function (built in functions from coordinator and functions from sidecar), then we would need to resolve naming collisions by only choosing one implementation when there exists a coordinator implementation and a worker implementation.
+
+* If we choose to have separate namespaces for built in functions from coordinator and functions from sidecar, then this would require extra changes on the worker code. Currently the worker code is using one prefix for every function, and this prefix is configurable. However, you cannot set a different prefix for different functions coming from the sidecar; they would all be the same prefix.
+
+* Since adding more worker changes adds extra complexity and confusion, we decided to use presto.default. prefix for every function from sidecar that is registered.
```
aditi-pandit, line 49: But then how are you handling conflicts then ? I do see it depends on the code you have suggested below. But can you describe what the desired outcome is in a functional manner here ?

RFC-0012-built-in-native-function-namespace-manager.md
kevintang2022: This RFC is about picking which FunctionHandle to use. If there is a SQL implementation and a C++ implementation, then the C++ function handle will be used. If there is a Java implementation, then the java function handle will take priority. By doing this,  it creates the side effect of which functions are constant folded or not, and this is determined by the implementation type.

This code ponter shows that Java implemenetation are constant folded, but C++ implementations are not.
https://github.com/prestodb/presto/blob/8041eb5e3ae0dc6e6a656d3e68d90ed8e1fda099/presto-spi/src/main/java/com/facebook/presto/spi/function/FunctionImplementationType.java#L16

PR #42, opened 2025-08-07
Coordinator throttling RFC

prashantgolash:
RFC for coordinator throttling.


RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,117 @@
...
+- **Cons**
+Extra service to maintain
+Two brains
+
+
+## Adoption Plan
+If this feature is enabled, this can cause queuing when cluster is loaded (which would be separate from queuing due to Resource Groups)
+. For now end users can correlate queuing via looking at queuing + worker overload metrics together. If required, we can also expose separate queuing metrics due to cluster overload.
+## Test Plan
+Added the test cases
```
rschlussel, line 116: Spell this out more about what kinds of tests you are adding.

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,117 @@
...
+- **Pros**
+No coordinator changes
+
+- **Cons**
+Extra service to maintain
+Two brains
+
+
+## Adoption Plan
+If this feature is enabled, this can cause queuing when cluster is loaded (which would be separate from queuing due to Resource Groups)
```
rschlussel, line 113:
```suggestion
Enabling this feature can cause additional queuing when the cluster is overloaded (which would be separate from queuing due to Resource Groups).
```

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,117 @@
...
+. For now end users can correlate queuing via looking at queuing + worker overload metrics together. If required, we can also expose separate queuing metrics due to cluster overload.
```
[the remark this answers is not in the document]
prashantgolash: Sounds good. I will add these metrics. Updated the section as well.

RFC-0011-coordinator-throttling.md
prashantgolash: Added the details on testing.

rschlussel (approved):

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,117 @@
...
+- PR: https://github.com/prestodb/presto/pull/25686
+
+### Coordinator side changes
+#### Collecting worker load data
+  - Update DiscoveryManager to invoke this end point rather than the `/v1/info/state`
+  - Expose `getNodeLoadMetrics` on InternalNodeManager
+  - PR: https://github.com/prestodb/presto/pull/25688 
+#### Scheduling policies
+  - Implement node overload policies that can govern the policies to determine cluster overload
+  - Based on these policies and load collected from worker, improve the admission control logic globally to not queue the query (irrespective of Resource Group)
```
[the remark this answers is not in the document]
tdcmeehan: Some folks rely on the JMX connector to monitor metrics.  I think there should be flexibility on how to define these limits so that we can separate them out with a resource group configuration (for example, it's a common resource group configuration to separate out DDL statements from other queries, precisely because DDLs don't use worker resources).

A good example is `ResourceGroupQueryLimits`.  This is a per-query limit that is defined at the RG level.  I could see this throttling specification being configured similarly.

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,134 @@
...
+
+For immediate relief, we propose implementing a holistic admission control mechanism that provides reactive throttling of the overall workload by not admitting queries.
+
+### Goals
+- Decrease duration of overload on the cluster
+- Also decrease overall variance of execution time of queries (queuing is fine)
+
+## Proposed Implementation
+### Worker side changes
+- Added endpoint to gather worker load (`/v1/info/nodestats`)
```
tdcmeehan, line 60: Can you include what is returned from this endpoint?

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,134 @@
...
+- PR: https://github.com/prestodb/presto/pull/25686
+
+### Coordinator side changes
+#### Collecting worker load data
+  - Update DiscoveryManager to invoke this end point rather than the `/v1/info/state`
+  - Expose `getNodeLoadMetrics` on InternalNodeManager
+  - PR: https://github.com/prestodb/presto/pull/25688 
+#### Scheduling policies
+  - Implement node overload policies that can govern the policies to determine cluster overload
+  - Based on these policies and load collected from worker, improve the admission control logic globally to queue the query (irrespective of Resource Group)
```
tdcmeehan, line 71: There isn't a whole lot of detail here.  Please be more specific--how is overload determined?

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,134 @@
...
+  - PR: https://github.com/prestodb/presto/pull/25689 
+
+
+## Metrics
+
+- Cluster overload
```
tdcmeehan, line 77: Is this a binary metric?  Can you elaborate?

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,134 @@
...
+- Variance in runtime of queries
```
tdcmeehan, line 78: Where and how is this calculated?

tdcmeehan (changes requested):
It seems like one flaw in this design is it's fundamentally unfair, meaning, one user can overwhelm a cluster and cause it to be unavailable for other multitenant users.  Have you thought about how to make the design more fair?  Resource groups are designed to provide some degree of fairness, have you considered a solution that groups load by resource group in order to throttle the problematic users while allowing less problematic users to continue to use the cluster?

RFC-0011-coordinator-throttling.md
spershin: @tdcmeehan 
Each worker is determining if it is overloaded according to the code and the config parameters.
Coordinator is merely notifier is the worker has memory or CPU overload at the time of reporting.

The worker code to determine the overload can now be thought of being version 1.
We think to improve it as it is not 'snappy' enough.

We could create some documentation and link to it here.
Would need advice where we can create a doc page describing worker overload detection and pushback and how verbose it should be.

RFC-0011-coordinator-throttling.md
spershin: +1

It would be good to document the fields (string constants) here or in the "## Metrics" section and refer to it.

prashantgolash: Added the response and field

RFC-0011-coordinator-throttling.md
prashantgolash: Added some details, but as @spershin  mentioned the majority of load logic is in cpp worker. I added link to committed PRs, but happy to update doc / create one.

RFC-0011-coordinator-throttling.md
prashantgolash: Added details on the JMX metrics added.

prashantgolash:
> It seems like one flaw in this design is it's fundamentally unfair, meaning, one user can overwhelm a cluster and cause it to be unavailable for other multitenant users. Have you thought about how to make the design more fair? Resource groups are designed to provide some degree of fairness, have you considered a solution that groups load by resource group in order to throttle the problematic users while allowing less problematic users to continue to use the cluster?

Initially I also explored enhancing RG level throttling mechanism, but there are some reasons I choose this approach.

Accuracy of Resource Group (RG) Resource Accounting:  
   The softMemoryLimits setting does not account for "unaccounted memory" that may be shared across multiple queries. As a result, accurately attributing actual worker resource usage to individual queries—and therefore to corresponding Resource Groups—can be imprecise. On the CPU side, I am also unsure whether throttling is implemented correctly (see [reference](https://github.com/prestodb/presto/blob/master/presto-main-base/src/main/java/com/facebook/presto/execution/resourceGroups/InternalResourceGroup.java#L1055)). In the future, we plan to introduce additional load metrics (e.g., queued drivers, I/O throttling) that may not align directly with the current RG configuration.
   
Granularity of RG Accounting:  
   At present, RG accounting operates at the stage->query level across the entire cluster. For our approach, we require more granular, worker-level resource utilization metrics. While it is technically possible to enhance the logic to map worker resource usage back to RGs, the limitations described in point (1) mean that this alone may not provide sufficient accuracy.
   
Lack of Tenant Isolation at the Worker Level:  
   Another key concern is tenant isolation. Admitting queries from users that are not causing issues may still result in those queries being routed to the same overloaded node. Since there is currently no tenant isolation at the worker level, simply admitting "non-problematic" queries is unlikely to resolve potential resource contention.

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,154 @@
...
+While not exhaustive, current throttling/possible mechanisms relevant for this discussion
+
+#### Granular task based scheduling
+
+**Leaf Task scheduling**
+
+The node selection logic in Presto involves randomly selecting 10 nodes from those that do not currently have scheduled splits for the given query. Among these 10 nodes, the system then chooses the node that is the least busy. For reference, the relevant code implementation can be found [here](https://github.com/prestodb/presto/blob/master/presto-main-base/src/main/java/com/facebook/presto/execution/scheduler/nodeSelection/SimpleNodeSelector.java#L182).
+Additionally, in the process of leaf task scheduling, a feedback mechanism is employed whereby splits are assigned to the nodes deemed least busy. It is important to note that in this context, "busyness" is defined by the number of splits being processed by a node, rather than its actual system load. 
+
+While this approach is generally effective for distributing work, it may not always result in the most optimal allocation of resources.
```
tdcmeehan, line 39: Can you elaborate on the difference between "effectiveness when distributing work" vs. "allocating resources"?  To me they sound the same--resources are worker resources, so effective distribution of work means effective allocation of resources.

RFC-0011-coordinator-throttling.md
tdcmeehan: Rather than a binary overloaded/not-overloaded, have you considered reporting raw metrics and allowing a user to self-determine what is a sensible overloaded value (with a sensible default)?

I am thinking that it would first make sense to queue on traditional resource group metrics, as these can have finer grained control and can isolate problematic users.  You would use this as a backstop in case that were not effective.  But they should be defined in conjunction with the resource group metrics, as the overload decision should involve metrics that exceed whatever was configured in the resource group.  For example, memory limits only rely on what is reported--for unreported memory, we use this as a backstop.

By the way, what sort of unreported memory is being allocated in Velox?

RFC-0011-coordinator-throttling.md
prashantgolash: In our scenario even with this logic of choosing least busy nodes based on splitting, most of the workers could become overload. I didn't meant to say this logic is wrong. Rather over-admitting is the key issue here.

RFC-0011-coordinator-throttling.md
prashantgolash: > I am thinking that it would first make sense to queue on traditional resource group metrics, as these can have finer grained control and can isolate problematic users.

Do you mean tune current RG configs to not let cluster overload to begin with?

> defined in conjunction with the resource group metrics, as the overload decision should involve metrics that exceed whatever was configured in the resource group. 
Not sure if I understood this part. Can you provide more details?

> By the way, what sort of unreported memory is being allocated in Velox?

Velox has different mechanism than java for memory allocation. I think RSS + cached memory is unaccounted for but @spershin may provide more details.

prashantgolash:
> > The softMemoryLimits setting does not account for "unaccounted memory" that may be shared across multiple queries.
> 
> To be clear, I'm not disagreeing with you that the existing metrics might not be sufficient to prevent overadmission, I'm simply pointing out that the current mechanism of queueing lacks fairness, which is not true for existing resource groups.
> 
> > As a result, accurately attributing actual worker resource usage to individual queries—and therefore to corresponding Resource Groups—can be imprecise.
> 
> What I proposed earlier would be a per-query limit on worker-reported total memory usage. I just think that the decision to queue should be _configured_ at a per-query level in the coordinator if we can't reliably aggregate these metrics from the task level.
> 
> > On the CPU side, I am also unsure whether throttling is implemented correctly (see [reference](https://github.com/prestodb/presto/blob/master/presto-main-base/src/main/java/com/facebook/presto/execution/resourceGroups/InternalResourceGroup.java#L1055)). In the future, we plan to introduce additional worker load metrics (e.g., queued drivers, I/O throttling) that may not align directly with the current RG configuration.
> 
> Queued drivers and IO all sound like they could be aggregated from the task level, which would make them eligible to be configured as new metrics in the resource group. This design leaves them as worker-determined binary flags. I would instead report the raw metrics, which could then be aggregated into resource groups and be configured far more flexibly.
> 
> > At present, RG accounting operates at the stage->query level across the entire cluster. For our approach, we require more granular, worker-level resource utilization metrics. While it is technically possible to enhance the logic to map worker resource usage back to RGs, the limitations described in point (1) mean that this alone may not provide sufficient accuracy.
> 
> For metrics that can be aggregated, they should be added as metrics for queueing consideration in the resource group. For metrics which can't be aggregated, I would recommend this decision be made on a per-query basis and configurable (for example, splitting DDL statements from execution, potentially classifying queries which only access connectors which use a single connection like JDBC, and splitting them from resource-heavy queries like Hive and Iceberg).
> 
> > Another key concern is tenant isolation. Admitting queries from users that are not causing issues may still result in those queries being routed to the same overloaded node. Since there is currently no tenant isolation at the worker level, simply admitting "non-problematic" queries is unlikely to resolve potential resource contention.
> 
> I would recommend that that also be addressed in this design. The Presto scheduler already uses heuristics to decide which nodes are eligible to be scheduled to. If the worker now is reporting reliable statistics to aid in this decision, then the scheduler should use this information to improve resource utilization and prevent this scenario. See `SimpleNodeSelector`.



> > The softMemoryLimits setting does not account for "unaccounted memory" that may be shared across multiple queries.
> 
> To be clear, I'm not disagreeing with you that the existing metrics might not be sufficient to prevent overadmission, I'm simply pointing out that the current mechanism of queueing lacks fairness, which is not true for existing resource groups.
> 
> > As a result, accurately attributing actual worker resource usage to individual queries—and therefore to corresponding Resource Groups—can be imprecise.
> 
> What I proposed earlier would be a per-query limit on worker-reported total memory usage. I just think that the decision to queue should be _configured_ at a per-query level in the coordinator if we can't reliably aggregate these metrics from the task level.
> 
Are you suggesting that if we know before hand that worker is running hot, we should not admit heavy queries. What would be the config look like per-query ? Would this be RG specific config?

> > On the CPU side, I am also unsure whether throttling is implemented correctly (see [reference](https://github.com/prestodb/presto/blob/master/presto-main-base/src/main/java/com/facebook/presto/execution/resourceGroups/InternalResourceGroup.java#L1055)). In the future, we plan to introduce additional worker load metrics (e.g., queued drivers, I/O throttling) that may not align directly with the current RG configuration.
> 
> Queued drivers and IO all sound like they could be aggregated from the task level, which would make them eligible to be configured as new metrics in the resource group. This design leaves them as worker-determined binary flags. I would instead report the raw metrics, which could then be aggregated into resource groups and be configured far more flexibly.
> 
> > At present, RG accounting operates at the stage->query level across the entire cluster. For our approach, we require more granular, worker-level resource utilization metrics. While it is technically possible to enhance the logic to map worker resource usage back to RGs, the limitations described in point (1) mean that this alone may not provide sufficient accuracy.
> 
> For metrics that can be aggregated, they should be added as metrics for queueing consideration in the resource group. For metrics which can't be aggregated, I would recommend this decision be made on a per-query basis and configurable (for example, splitting DDL statements from execution, potentially classifying queries which only access connectors which use a single connection like JDBC, and splitting them from resource-heavy queries like Hive and Iceberg).
> 
> > Another key concern is tenant isolation. Admitting queries from users that are not causing issues may still result in those queries being routed to the same overloaded node. Since there is currently no tenant isolation at the worker level, simply admitting "non-problematic" queries is unlikely to resolve potential resource contention.
> 
> I would recommend that that also be addressed in this design. The Presto scheduler already uses heuristics to decide which nodes are eligible to be scheduled to. If the worker now is reporting reliable statistics to aid in this decision, then the scheduler should use this information to improve resource utilization and prevent this scenario. See `SimpleNodeSelector`.

RFC-0011-coordinator-throttling.md
spershin: @tdcmeehan 

Thanks for the comments!

We have some (I would call them) blockers for making this framework tied to RGs at the moment.
I'll try to describe them here.

1. In the worker we do not have a good understanding of how much memory each query uses due to the untraceable memory (this cannot be fixed now).
The cpu usage snapshot (not total so far) would be also hard if not impossible to collect per query, although we could measure something like how many active execution threads each query has. This might be good to consider in the long run.

2. The idea of not allowing queries to be started from the coordinator is an extension of the worker pushback framework. The latter is just each worker postpones starting task if the worker is overloaded. This creates certain ambiguities for query metrics: this postponing time contributes to the execution time rather than queued time, the different queries might be queued in different order on different workers, etc.
To help with it we notify coordinator about the overload, so we can have a central entity queuing the majority of the queries during the overload times.
So, it is not so much as for the coordinator to manage it at the moment, although in the future we might do something like that.

3. RGs are not fair at all - they do not really consider CPU and memory most of the time: memory logic is there, but since we cannot report proper memory from the worker it is not really reflecting the situation. In other words it is mainly based on the number of queries and we know that all queries aren't equal.

4. Queuing ALL queries can be considered fair if we don't look at priorities.

We propose to add the suggestion to the RFC as the potential to improve the framework and continue with the planned implementation for the phase 1.

As for the other and more detailed metrics - it is possible. We have map<string, double> in the endpoint structure and it can have anything we want in the future. For the phase 1 we thought to keep it simple.

RFC-0011-coordinator-throttling.md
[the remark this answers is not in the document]
tdcmeehan: DDLs and system queries too.  For example, if I need to query the information_schema table, it should not queue at all--none of that is going to the workers, it's just metadata.

RFC-0011-coordinator-throttling.md
prashantgolash: @tdcmeehan  - I have added sections for

1. Using raw metrics
2. Improvement as extension of this RFC (In Approach-1)
3. Query configs to throttle at query level

Let me know if this looks good to you

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,172 @@
...
+For immediate relief, we propose implementing a holistic admission control mechanism that provides reactive throttling of the overall workload by not admitting queries.
+
+### Goals
+- Decrease duration of overload on the cluster
+- Also decrease overall variance of execution time of queries (queuing is fine)
+
+## Proposed Implementation
+### Worker side changes
+- Added endpoint to gather worker load (`/v1/info/nodestats`)
+  - Include NodeState + worker load. Idea is to just call `/v1/info/nodestats` rather than `/v1/info/state` (eventually deprecate it). Worker load is already getting populated in [cpp workers](https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L1537).
```
tdcmeehan, line 61: Can this plan include the deprecation?  I think we need to clean up the old endpoint as well so we're not confusingly returning two different things depending on whether or not we're running a C++ cluster.

tdcmeehan (commented):
If a worker has self-identified as overloaded, would its corresponding `NodeState` be updated as `INACTIVE`/`SHUTTING_DOWN`?

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,172 @@
...
+**Intermediate task scheduling**
+
+For intermediate task scheduling in Presto, the selection of workers is based on a hash partition count. Specifically, intermediate tasks are assigned to worker nodes according to the hash of the partition, which determines the distribution across available intermediate workers. The relevant code implementation can be found [here](https://github.com/prestodb/presto/blob/master/presto-main-base/src/main/java/com/facebook/presto/sql/planner/SystemPartitioningHandle.java#L164).
+This strategy does not consider the current system load when assigning intermediate tasks. 
+
+#### Resource Group throttling
+There are certain metrics like softMemoryLimit etc which exist at RG level to queue the workload, but it does not consider the actual utilization and is out of scope for this discussion.
+
+
+While we could improve task scheduling (pro-active measure to not let worker overload) through load balancing (more on this in alternative approaches), this alone would be insufficient given the nature of our overload scenario where all workers are saturated in short period of time which is due to admitting lot of heavy queries in short duration of time.
```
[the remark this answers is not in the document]
prashantgolash: I was referring to the task scheduling. I have updated the description to make it accurate.

RFC-0011-coordinator-throttling.md
prashantgolash: Added the deprecation plan

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,154 @@
...
+## Proposed Implementation
+### Worker side changes
+- Added endpoint to gather worker load (`/v1/info/nodestats`)
+  - Include NodeState + worker load. Idea is to just call `/v1/info/nodestats` rather than `/v1/info/state` (eventually deprecate it). Worker load is already getting populated in [cpp workers](https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L1537).
+  - **Fields**
+    - **nodeState** [Code link](https://github.com/prestodb/presto/blob/master/presto-spi/src/main/java/com/facebook/presto/spi/NodeState.java)  
+    - **metrics**
+      - Map<String, Double> loadMetrics
+        - key is overload state like `cpu.overload`, `memory.overload`
+        - val is 0 or 1 to indicate overload or not
```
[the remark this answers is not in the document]
prashantgolash: Yes, will make it an object as updated in the newer version

RFC-0011-coordinator-throttling.md
```diff
@@ -0,0 +1,180 @@
...
+
+- **Pros**
+  - RG throttling code could be reused
+  - More fair to throttle workloads which are actually overloading
+
+- **Cons**
+  - Needs significant work on worker side
+  - Also need work on coordinator side
+  - Need more analysis on key metrics
+  - No enforcement: Even if we fix the accounting and improve admission control to run the cluster less loaded, there is no guarantee that RGs won't exceed their limits, potentially causing some workers to become overloaded since Presto lacks preemption support and queries can use more resources once admitted.
```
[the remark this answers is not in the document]
prashantgolash: Yes you are right - we are not enforcing in this model either. I only brought this up in context of RG accounting fairness guarantees. I am ok framing current RFC as second line of support mechanism. Do you recommend I remove RG from alternative approach section and update the `Background` section with these details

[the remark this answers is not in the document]
prashantgolash: Updated the RFC per this comment.

spershin (approved):
Looks good!
Thank you!

prashantgolash:
> @prashantgolash can you please fix up the numbering to reflect the latest number in this repo? Thanks

Done

tdcmeehan (approved):

PR #50, opened 2025-10-13
Add RFC to support custom schemas in native sidecar function registry

Joe-Abraham:


tdcmeehan (approved):
LGTM, just one question

pdabre12 (commented):
Thanks @Joe-Abraham .
Please change the RFC title to this `Add RFC to support custom schemas in native sidecar function registry` or something similar.

RFC-0017-Support-custom-schemas-in-native-sidecar-function-registry.md
```diff
@@ -0,0 +1,69 @@
...
+- Functions are registered in Velox as `catalog.schema.function_name`.
+- Filtering is performed by splitting the function name and matching the catalog.
+
+## Usage Example
+Configure multiple function namespace managers:
+```properties
+# etc/function-namespace/mycatalog.properties
+function-namespace-manager.name=native
+function-implementation-type=CPP
+supported-function-languages=CPP
```
aditi-pandit: What do these 2 fields mean and what are the possible options for them ?

aditi-pandit (approved):

RFC-0017-Support-custom-schemas-in-native-sidecar-function-registry.md
Joe-Abraham: `function-implementation-type` - Indicates the language in which functions in this namespace are implemented. And it can take the following [values](https://github.com/prestodb/presto/blob/master/presto-spi/src/main/java/com/facebook/presto/spi/function/FunctionImplementationType.java#L17-L23)

`supported-function-languages` - Languages supported by the namespace manager. 
I don't see any pre-list of values this config could take, In current codebase `cpp`, `sql` and `java` are used

Joe-Abraham:
@kevintang2022 
> How does the function resolution handle the case when the native worker registers functions under a namespace and the coordinator registers functions under the same function namespace? I wonder if only one function catalog can be mapped to exactly one namespace manager.

Only one function catalog can map to one namespace manager, native worker functions use a parallel system through BuiltInFunctionHandle with different kinds to coexist with coordinator functions without catalog collision.

pdabre12 (approved):
Thanks @Joe-Abraham

tdcmeehan (approved):

PR #52, opened 2025-11-11
Add GPU RFC

majetideepak:


tdcmeehan:
Can we add some thinking around the optimizer?  It would seem that the plan might differ between CPU based execution and GPU, for example, I imagine that regular expressions might often be more efficient when executed on a CPU?  And does task parallelism change between CPU and GPU based execution, and if so, wouldn't that mean we need to alter the scheduler as well?  Finally, if we will support both CPU and GPU based computation side by side, is there any thinking about admission control and how to tune that to optimize mixed execution?

PR #53, opened 2025-11-25
Add RFC for supporting MCP with Presto

agrawalreetika:
Add RFC for supporting MCP with Presto


RFC-0022-Model-Context-Protocol.md
```diff
@@ -0,0 +1,127 @@
...
+
+Introduce a new lightweight service: presto-mcp-server, deployed alongside Presto coordinators and Presto Router.
+
+The MCP server will:
+
+- Expose a JSON-RPC 2.0 HTTP endpoint (`/mcp`, `/v1/mcp`) interface to AI clients.
+- Implement the core MCP primitives:
+  - `tools/list` for tool discovery 
+  - `tools/call` for executing tools
+  - Initially provide a single tool: `query.run`, which executes SQL queries against Presto.
```
tdcmeehan: Any restriction on the type of queries supported?  If it's a DML, might we need a new tool to support it, since they don't return data but return metadata about what was modified?

RFC-0022-Model-Context-Protocol.md
```diff
@@ -0,0 +1,127 @@
...
+- Internally communicate with Presto coordinators using standard Presto HTTP APIs.
+- Forward OAuth/JWT Bearer tokens transparently from MCP clients to Presto, ensuring that Presto performs all authentication and authorization checks.
+- Translate between the two protocols, aggregating streaming results into a single response.
```
tdcmeehan: What if the query requests too much data and creates memory pressure in the MCP server?

RFC-0022-Model-Context-Protocol.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Final Thoughts
+
+This proposal cleanly bridges Presto with next-generation agent ecosystems (LLMs, AI workflows, model interaction tools). The MCP server architecture respects Presto’s deployment patterns, is backward-compatible, and provides a robust extension point for future interactive functionality such as:
+
+- Schema browsing tools
+- Table metadata tools
+- Query explanation tools
+
+## WIP - Draft PR Changes
```
tdcmeehan:
```suggestion
```

I presume it's not WIP anymore?

RFC-0022-Model-Context-Protocol.md
aditi-pandit: I presume this will also run SHOW catalogs; and other statements ?

agrawalreetika: Yes, `query.run` in the initial version is intentionally limited to read-only, row-producing queries (e.g., SELECT, SHOW, DESCRIBE).
Then there would be additional metadata oriented different tools - `metadata.listCatalogs`, `metadata.listSchemas`, `metadata.listTables`, `metadata.getColumns`

Thanks for pointing out, will update RFC details as well.

agrawalreetika: Yes — `query.run` is intentionally limited to read-only, row-producing queries.
LLM agents primarily issue analytical (SELECT-style) queries, and MCP tools are designed to return tabular JSON data.
My understanding is that if support for mutations is needed later, they could be exposed as separate MCP tools with their own semantics.

RFC-0022-Model-Context-Protocol.md
agrawalreetika: Good point - Although results are aggregated, the MCP server enforces a strict row limit (auto-injected if missing), ensuring memory is bounded and preventing large result sets from being materialized.

I am open to suggestions if additional safeguards should be added.

hantangwangd (commented):
@agrawalreetika thanks for this interesting proposal. Is there a prototype or draft PR already in the works that we could look at?

PurpleSquirrelMedia (commented):
## Review from MCP Server Developer

Thanks for the RFC @agrawalreetika! As someone who's built 9 production MCP servers with 200+ tools across various platforms (IBM Cloud, Oracle Cloud, DaVinci Resolve, etc.), I wanted to share some feedback based on real-world implementation experience.

### What's Excellent

1. **Architecture Decision**: The separate `presto-mcp-server` process is the right call. This mirrors how most production MCP servers work - a lightweight translation layer that doesn't pollute the core system.

2. **Stateless Design**: Delegating all state to Presto coordinators is clean and matches MCP's design philosophy.

3. **Tool Taxonomy**: The `query.run` + `metadata.*` structure is well-organized. Consider namespacing as `presto_query_run`, `presto_list_catalogs` etc. for clarity when multiple MCP servers are configured.

4. **Security**: Token forwarding approach is solid.

### Suggestions for Enhancement

#### 1. **Add Streaming Progress Notifications**
For long-running queries, consider implementing MCP's `notifications/progress` capability:

```json
{
  "jsonrpc": "2.0",
  "method": "notifications/progress",
  "params": {
    "progressToken": "query-123",
    "progress": 45,
    "total": 100
  }
}
```

This would map nicely to Presto's `stats` in the query response (completedSplits/totalSplits).

#### 2. **Consider MCP Resources Endpoint**
Beyond tools, MCP supports `resources` for data discovery:

```json
{
  "method": "resources/list",
  "result": {
    "resources": [
      {
        "uri": "presto://catalog/schema/table",
        "name": "users",
        "mimeType": "application/json"
      }
    ]
  }
}
```

This could expose catalogs/schemas/tables as browsable resources.

#### 3. **Pagination for Metadata**
The `metadata.listTables` could return thousands of tables. Consider:
- `cursor` parameter for pagination
- `limit` parameter (default 100)
- Return `nextCursor` in response

#### 4. **Prompts Capability**
MCP's `prompts` feature could provide templates:

```json
{
  "name": "top_tables_by_size",
  "description": "Show largest tables in a schema",
  "arguments": [{"name": "schema", "required": true}]
}
```

#### 5. **Error Handling Details**
Recommend mapping Presto error codes to MCP error codes:
- `PERMISSION_DENIED` → -32001
- `QUERY_REJECTED` → -32002
- `SYNTAX_ERROR` → -32602 (Invalid params)

#### 6. **Configuration Considerations**
From my experience, useful config options:
- `mcp.query.max-rows`: Default result limit
- `mcp.query.timeout-seconds`: Query timeout
- `mcp.metadata.cache-ttl`: Cache catalog/schema metadata

### Implementation Offer

I'd be happy to:
1. Provide code samples from my existing MCP servers
2. Review the implementation PR
3. Test with Claude/GPT integrations

My relevant repos for reference:
- https://github.com/PurpleSquirrelMedia/ibmcloud-mcp-server (80+ tools)
- https://github.com/PurpleSquirrelMedia/oracle-mcp-server (21 tools, includes DB operations)

Great work on this RFC! 🎉

PR #55, opened 2026-01-07
Add presto jvector integration RFC

NivinCS:


tdcmeehan:
@NivinCS it's not clear from this document which interfaces go into the SPI.  Can you please elaborate on this design from the perspective of the plugin author: what SPI interfaces need to be implemented, and also how does this SPI integrate into the engine?  It simply misses this foundational detail.

gggrace14:
Hi @NivinCS , it's not easy for readers to tell what will be added/changed to each component, the syntax, analyzer, SPI, planner, operator, function, etc. Do you think it'll help if you organize by component after introducing the overall logic flow?

NivinCS:
> @NivinCS it's not clear from this document which interfaces go into the SPI. Can you please elaborate on this design from the perspective of the plugin author: what SPI interfaces need to be implemented, and also how does this SPI integrate into the engine? It simply misses this foundational detail.

Hi Tim,
We’ve updated the RFC to better explain which SPI interfaces a connector must implement and how those interfaces are integrated into the engine, covering both index building and vector search . Could you please take a look and let us know your review comments ?

mohsaka:
@NivinCS 
The approx_nearest_neighbors table function looks fine to me. Since it's not taking in any tables its a leaf table function operator and you should implement the TableFunctionSplitProcessor. 

I took a closer look at the TVF Implementation and now I understand what you meant by applying to partitions.

I got confused when it came to partitioning, but now I think I understand. The partition is actually the partition in the jvector not the partition of the table. Yes, we should create a split for each of these partitions via TableFunctionProcessorProvider. 

This all looks fine to me from a table function standpoint. Thanks!

mohsaka (commented):
Looks good from a table function standpoint.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1317 @@
...
+    // JSON serialization methods
+}
+```
+
+**TopKAggregationOperator**
+
+Coordinator-side operator that aggregates results from all workers to produce global top-K. either we can use k-way merge or bruteforce search to efficiently maintain the best K results.
+
+```java
+public class TopKAggregationOperator implements Operator {
```
aditi-pandit: Why  is this a new operator ? Can we achieve the same functionality with a custom aggregation function or another table operator (which is a data processor vs the previous split processor now)

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1317 @@
...
+
+Consider a document search system with embeddings:
+
+```sql
+-- Create table with embeddings
+CREATE TABLE documents (
+    doc_id BIGINT,
+    title VARCHAR,
+    content VARCHAR,
+    embedding ARRAY(REAL),  -- 768-dimensional embedding
```
aditi-pandit, line 61: I understand this is a column in the table, but can you give the SQL to insert data into this table ?

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1317 @@
...
+     * @return Page with global top-K results sorted by score
+     */
+    @Override
+    public Page getOutput();
+}
+```
+
+### Architecture Overview (index building flow)
+
+The implementation follows a **begin-execute-finish pattern** with distributed execution: once user calls `CREATE INDEX` statement, the coordinator validates parameters and generates splits for each partition. Each worker then processes its assigned splits to build the index. The coordinator aggregates results and finalizes the index creation.
```
aditi-pandit: Seems below that a CALL procedure is used for index building by the user not a CREATE INDEX command. Please can you clarify.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1317 @@
...
+        String indexName,
+        PartitionInfo droppedPartition);
+}
+```
+
+### Performance Optimizations
+
+#### 1. Index Caching
+
+**VectorIndexCache**
```
aditi-pandit, line 970: In which component will this be used ?

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1317 @@
...
+```
+
+#### Dynamic Partition Management
+
+**PartitionAwareIndexManager**
+
+Manages indexes for individual partitions, supporting incremental index updates as partitions are added or removed.
+
+```java
+public interface PartitionAwareIndexManager {
```
aditi-pandit, line 932: In which Presto components will this be used ?

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1317 @@
...
+        int k,
+        double targetRecall);
+}
+```
+
+#### 3. Parallel Index Building
+
+**ParallelIndexBuilder**
+
+Builds multiple partition indexes concurrently using Presto's distributed execution framework:
```
aditi-pandit: Since the index building happens concurrently on all workers the parallelism seemed implicit to me. Are you suggesting there that this will be used to introduce parallelism within the single worker itself ?

aditi-pandit (commented):
Thanks @NivinCS for this writeup. I have a bunch of questions.

Dilli-Babu-Godari:
> Hi @NivinCS , it's not easy for readers to tell what will be added/changed to each component, the syntax, analyzer, SPI, planner, operator, function, etc. Do you think it'll help if you organize by component after introducing the overall logic flow?

Hi @gggrace14, thanks for the suggestion. The design intentionally does not introduce changes to core components; the impact is limited to connector implementations only, which plug into existing engine behavior. We can add a short section to explicitly clarify the non-impact on other components if needed.

RFC-0022-jvector-integration.md
NivinCS: We introduced a new `TopKAggregationOperator` to merge pre-sorted results produced by workers. One limitation we observed with custom aggregation functions is that they process input rows in an arbitrary order, which causes the loss of the sorted order already produced by the workers.

Since the `TableFunctionDataProcessor` runs on workers and processes partitions independently, it cannot perform aggregation across results from all workers. A coordinator-side aggregation step is therefore required.

While Presto already provides result aggregation through the `TopNOperator`, it is primarily designed for general use cases. In our scenario, the `LIMIT` clause already constrains the result to `top-K`, and `TopNOperator` uses a heap-based algorithm that does not take advantage of pre-sorted inputs, making it less efficient.

RFC-0022-jvector-integration.md
NivinCS: added  a sample INSERT query in RFC including a realistic 768-dimension embedding example (shortened for readability)
```
INSERT INTO documents (
    doc_id,
    title,
    content,
    embedding,
    created_date
)
VALUES (
    1001,
    'Vector Search in Presto',
    'This document explains how vector similarity search works in Presto.',
    ARRAY[
        0.0123, -0.3456, 0.7891, 0.4567, -0.1123,
        0.9981, -0.2234, 0.3345, 0.6678, -0.5543
        -- ... continue until 768 REAL values
    ],
    DATE '2026-03-01'
);
```

RFC-0022-jvector-integration.md
NivinCS: Yes, you’re right. we invoke a CALL procedure, not a CREATE INDEX statement. updated in RFC.

RFC-0022-jvector-integration.md
NivinCS: The cache is connector-specific and lives entirely within the Iceberg connector implementation, making it a plugin-level optimisation rather than a core Presto feature.

RFC-0022-jvector-integration.md
NivinCS: in Iceberg connector only

RFC-0022-jvector-integration.md
NivinCS: You’re right. A ParallelIndexBuilder is not needed, as parallelism is implicit.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+      "vector-index-ef-construction": 100,
+      "vector-index-build-timestamp": 1704470400000
+    }
+  }]
+}
+```
+**Benefits:**
+- ✅ Versioning: Index tied to specific snapshot
+- ✅ Consistency: Validate index via snapshot ID comparison
+- ✅ Cleanup: Expired snapshots auto-clean index references
```
tdcmeehan, line 836: How do the vector mappings and indexes get cleaned up?

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+    private final String indexPath;
+    private final String mappingPath;
+    private final float[] queryVector;
+    private final long limit;
+    
+    // JSON serialization methods
+}
+```
+
+**TopKAggregationOperator**
```
tdcmeehan: I was under the impression that this RFC was purely at the connector/plugin layer, however if this is a new node, how exactly will it be planned and executed?  This does seem like a core Presto change.  Presumably this will be planned--how?  How does the planner know to insert this node?

tdcmeehan (changes requested):
I think there's a hidden presumption in this PR--that "row IDs" are stable.  To be clear, there is no concept of row ID in V1 tables, in V2 tables they do exist but are **not stable**, and only in V3 tables are row IDs stable.  By stable, I mean, do they survive things like compaction, snapshot expirations, branches, etc.  If there's a hidden dependency on Iceberg V3 table format, which has **stable row IDs**, we should call it out explicitly in this RFC.  If not, then we need to mention this as a potentially significant limitation in this design.  It's significant because a compaction will wipe out the indices.  Likewise, if the table is copy-on-write, a single update or delete will at best wipe out the indices, at worst we'll return stale data.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+     * @param tableLayoutHandle Handle to the target table
+     * @param arguments Procedure arguments
+     * @return BuildVectorIndexHandle with execution context
+     */
+    @Override
+    public ConnectorDistributedProcedureHandle beginCallDistributedProcedure(
+        ConnectorSession session,
+        QualifiedObjectName procedureName,
+        ConnectorTableLayoutHandle tableLayoutHandle,
+        List<Object> arguments);
```
hantangwangd: It's not entirely clear to me about one point here. Are you suggesting we implement a new `IcebergMetadata` specifically for vector indices building and searching? Or, in other words, is the goal here to introduce a new Iceberg catalog type dedicated to vector indexing? If that's not the case, then I believe we likely don't need to re-implement these two methods. The actual logic should be handled within `BuildVectorIndexDistributedProcedure`. What are your thoughts on this?

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+     * @param layout Contains VectorSearchHandle from analysis
+     * @param splitSchedulingContext Scheduling context
+     * @return ConnectorSplitSource producing VectorSearchSplit objects
+     */
+    @Override
+    public ConnectorSplitSource getSplits(
+        ConnectorTransactionHandle transaction,
+        ConnectorSession session,
+        ConnectorTableLayoutHandle layout,
+        SplitSchedulingContext splitSchedulingContext);
```
hantangwangd: Is it intentional to override this method rather than the following table function specific method?

```
@Override
public ConnectorSplitSource getSplits(
            ConnectorTransactionHandle transaction,
            ConnectorSession session,
            ConnectorTableFunctionHandle function)
```

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+
+4. **ConnectorPageSinkProvider**:
+   - Creates page sinks for workers to write index build results
+   - Workers write fragments (index metadata) back to coordinator
+   - Handles serialization of build results (index paths, statistics)
+
+5. **ConnectorSplitManager**:
+   - Generates splits for parallel index building
+   - Creates one split per partition for partitioned tables
+   - Each split contains partition-specific information (data location, vector column, parameters)
```
hantangwangd: Is the detailed explanation for this part missing below?

hantangwangd (commented):
Thanks for this great feature.

Before describing the ANN search flow and index building flow in detail, it might be beneficial to begin by introducing the concept of partitions as used in this proposal, clarifying their relationship to both Iceberg table partitions and index partitions, and outlining the philosophy behind partition-based distributed execution. This would provide crucial context to readers before diving into the two flows.

RFC-0022-jvector-integration.md
aditi-pandit: Having such a column looks hard to use and also could have problems as we would want embeddings to be maintained as data changes.

There seem to be special vectorizer kind of components for this https://github.com/timescale/pgai/blob/main/docs/vectorizer/overview.md

We should explore something on those ideas for this work.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+     * Return the final global top-K results.
+     *
+     * @return Page with global top-K results sorted by score
+     */
+    @Override
+    public Page getOutput();
+}
+```
+
+### Architecture Overview (index building flow)
```
aditi-pandit: The document would be more readable if you added about the index building before the query part as the index build is a key part of the user workflow.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+    'This document explains how vector similarity search works in Presto.',
+    ARRAY[
+        0.0123, -0.3456, 0.7891, 0.4567, -0.1123,
+        0.9981, -0.2234, 0.3345, 0.6678, -0.5543
+        -- ... continue until 768 REAL values
+    ],
+    DATE '2026-03-01'
+);
+
+-- Find top 10 most similar documents to a query
```
aditi-pandit, line 97: Since the user is expected to create an index as well, it would be good to add those SQL here also.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+
+Consider a document search system with embeddings:
+
+```sql
+-- Create table with embeddings
+CREATE TABLE documents (
+    doc_id BIGINT,
+    title VARCHAR,
+    content VARCHAR,
+    embedding ARRAY(REAL),  -- 768-dimensional embedding
```
aditi-pandit, line 61: Also most vector search implementations in databases/query engines are making the embedding a special vector type. Is that something you have considered ? If yes, then what would be the implications to this proposal.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+```java
+@Description("Approximate Nearest Neighbors search for vector similarity")
+public class ApproxNearestNeighborsFunction extends AbstractConnectorTableFunction {
+    
+    /**
+     * Constructor declares function signature:
+     * - Arguments: query_vector (ARRAY(REAL)), column_name (VARCHAR), limit (BIGINT)
+     * - Return type: TABLE(row_id BIGINT, score REAL)
+     */
+    public ApproxNearestNeighborsFunction() { ... }
```
aditi-pandit: Can you give more details... Seems like the return type will be DescribedTableReturnTypeSpecification. 

The input parameters also don't seem to include a TABLE argument which is fine.

But if would be good if you wrote more specifics here.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+     * Analyze the table function call and return output schema.
+     * Validates arguments, loads index metadata, creates execution handle.
+     *
+     * @param session The connector session
+     * @param transaction The transaction handle
+     * @param arguments Function arguments (query_vector, column_name, limit)
+     * @return TableFunctionAnalysis with output schema and VectorSearchHandle
+     */
+    @Override
+    public TableFunctionAnalysis analyze(
```
aditi-pandit: The TableFunctionAnalysis typically returns a TableFunctionHandle. What will you include in it ?

I'm also a bit confused about how you are accessing data... In the sense you have a documents table with an embedding column and an index as well. The TableFunction analyze should typically do some validation or setup. This means you need to access the underlying table to validate the columns or presence of an index. How are you doing that ? Where is the metadata picked from ?

RFC-0022-jvector-integration.md
aditi-pandit: Its not possible to take advantage of sortedness unless the topK is co-located with the prior results from thw worker. If there is an exchange of the worker results then its not guaranteed that the topK operator is going to see sorted results from all the workers. The buffers between workers are mixed losing the sortedness. I don't think you can get by without using a heap-based algorithm like TopN.

Were you able to prototype this work with TopN but not happy with its performance ?

RFC-0022-jvector-integration.md
aditi-pandit: +1... I think to getSplits you will need information about the underlying documents table and index and those information will be passed in the ConnectorTableFunctionHandle.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+
+#### Index Creation
+
+```sql
+-- Create a vector index using stored procedure
+CALL system.create_vector_index(
+    table_name => 'catalog.schema.table_name',
+    column_name => 'embedding_column',
+    index_name => 'embedding_idx',
+    similarity_function => 'COSINE',  -- COSINE, EUCLIDEAN, DOT_PRODUCT (optional)
```
aditi-pandit, line 747: Does this presume a certain algorithm and similarity metric ? If we want to use some other index algorithm, then how would you repurpose this ?

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+);
+
+-- Drop an index
+CALL system.drop_vector_index(
+    table_name => 'catalog.schema.table_name',
+    index_name => 'embedding_idx'
+);
+
+-- Refresh an index (rebuild with latest data)
+CALL system.refresh_vector_index(
```
aditi-pandit, line 759: Indexes can get out of sync -- What happens behind the scenes here ? Will you be executing a DROP followed by CREATE ?

Other offerings seem to do https://docs.databricks.com/aws/en/vector-search/vector-search

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+}
+```
+
+### Performance Optimizations
+
+#### 1. Index Caching
+
+**VectorIndexCache**
+
+Worker-side cache for loaded indexes and mappings using any caching mechanism. Reduces repeated downloads from S3/HDFS.
```
aditi-pandit, line 972: Do you have any particular caching system in mind ? How do you ensure stickiness that subsequent queries use the same worker for the same split ?

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+
+```java
+public interface VectorIndexCache {
```
aditi-pandit, line 975: Which component is using this ? I'm not following where this fits in.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+     *
+     * @param indexPath Path to invalidate
+     */
+    void invalidate(String indexPath);
+}
+```
+
+#### 2. Adaptive Search Parameters
+
+**AdaptiveSearchParameterOptimizer**
```
aditi-pandit, line 1006: Where is this used ? Presume in split generation in table function ?

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1293 @@
...
+    doc_id BIGINT,
+    title VARCHAR,
+    content VARCHAR,
+    embedding ARRAY(REAL),  -- 768-dimensional embedding
+    created_date DATE
+) WITH (
+    partitioning = ARRAY['created_date']
+);
+
+-- sample INSERT query including a realistic 768-dimension embedding example (shortened for readability)
```
aditi-pandit, line 67: All the work here is described in context of iceberg... But if we use any other table format like delta, much of these SQL constructs are still applicable. 

Is there any particular reason to tie this implementation to iceberg ? Is it possible to separate this implementation into generic and specific pieces to wire between the different table formats ?

@gopikris @tdcmeehan

aditi-pandit (commented):
@NivinCS : Had another question about security related features as well ? Most of the Vector search libraries used for RAG also mention about how they provide secure access. Is that a concern at all in this work ?

RFC-0022-jvector-integration.md
NivinCS: Yes, we should be using ConnectorTableFunctionHandle. We’ve updated the RFC accordingly. Thank you for pointing this out.

RFC-0022-jvector-integration.md
NivinCS: During our initial prototype, we used Presto’s TopNOperator to compute the final Top-K results. This prototype ran on a single node with a relatively small dataset (around 10k vectors), and we did not observe any issues in that setup. Since the Jvector index  search returns results in sorted order, we considered introducing a customised operator to potentially achieve better performance.

However, as you pointed out, the sorted order from workers is not preserved during data transfer to the coordinator .We can continue to use the existing TopNOperator for global Top-K aggregation. This approach allows us to reuse existing functionality and avoids introducing any changes to Presto’s core components.

RFC-0022-jvector-integration.md
NivinCS: We had a similar discussion with @aditi-pandit as well. As mentioned in this [comment](https://github.com/prestodb/rfcs/pull/55#discussion_r2703578786) , we have updated the RFC to use the existing TopNOperator for global Top-K aggregation, thereby avoiding the introduction of any changes to Presto’s core components.

RFC-0022-jvector-integration.md
Dilli-Babu-Godari: You’re right. There’s no need to introduce a new `IcebergMetadata` implementation or a separate Iceberg catalog type for vector index building.

The index build logic can be handled entirely within `BuildVectorIndexDistributedProcedure`, with the required metadata updates performed during the finish phase.

Based on this, we don’t need to re-implement those metadata methods. I’ve updated the RFC accordingly. Thanks for calling this out.

RFC-0022-jvector-integration.md
Dilli-Babu-Godari: Thanks for suggesting this. Updated the RFC accordingly.

RFC-0022-jvector-integration.md
tdcmeehan: How does the planner know to insert the TopNOperator?

RFC-0022-jvector-integration.md
bibith4: Yes, we are currently presuming the index-building algorithm to be HNSW, since jvector only supports HNSW. However, users can still choose among different similarity metrics, such as cosine, Euclidean , or dot product. By default, we use cosine similarity

RFC-0022-jvector-integration.md
NivinCS: The Table-Valued Function (TVF) planning explicitly declares the required ordering and limit, so that the engine’s existing planning rules naturally introduce a TopN during plan generation. In our use case we applies ORDER BY ann.score LIMIT k on the Table-Valued Function output.

RFC-0022-jvector-integration.md
NivinCS: We did consider introducing a dedicated vector data type in both Iceberg and Presto early in the design. However, doing so would require changes across multiple layers: a new type definition in Iceberg, corresponding reader/writer and schema support, and engine-level type handling in Presto. This would likely require maintaining a temporary Iceberg fork until the type is accepted by the Iceberg community, significantly increasing delivery and maintenance cost.

Most databases introduce a vector type primarily to enable efficient vector operations and indexing. In our design, those capabilities are already provided by the JVector library itself, while Presto only needs a way to pass embeddings to the index. Representing embeddings as ARRAY<REAL> is sufficient for this purpose and avoids introducing new type semantics into the engine.

Additionally, Presto has historically been conservative about adding new primitive types unless there is a strong, engine-level justification. As discussed earlier (and as suggested by @tdcmeehan ), introducing a vector data type should be considered only if there is a clear requirement that cannot be met with existing types. At this stage, the proposal intentionally minimises scope by reusing ARRAY<REAL>, keeping the design simpler, more portable, and aligned with both Presto and Iceberg community expectations.

We see a dedicated vector type as possible future work if concrete limitations emerge that cannot be addressed with the current approach.

RFC-0022-jvector-integration.md
Dilli-Babu-Godari: I have updated the RFC with detailed information for this. Thanks you!

RFC-0022-jvector-integration.md
NivinCS: The AdaptiveSearchParameterOptimizer is used in the ANNSplitProcessor during split execution on worker nodes to dynamically calculate the optimal ef_search parameter for HNSW searches based on index size, K value, and target recall, though this is planned as a future enhancement with static configuration used initially.

RFC-0022-jvector-integration.md
NivinCS: The VectorIndexCache is used by the connector's TableFunctionSplitProcessor implementation (specifically ANNSplitProcessor) which is a Presto SPI component that executes on worker nodes during split execution - it calls getIndex() to load the vector index from cache before performing the JVector ANN search. (This can be accommodated as a future enhancement to provide better performance by reducing repeated downloads from S3/HDFS storage. )

RFC-0022-jvector-integration.md
NivinCS: We don't have a particular caching system in mind at this stage. During our internal discussions, this point came up and we thought it would be better to keep this as a future enhancement . We have a basic idea of the approach, but we need to explore more on the specific caching implementation and worker affinity mechanisms.

The VectorIndexCache interface is intentionally designed to be implementation-agnostic, allowing flexibility to choose the most appropriate caching strategy (e.g., Caffeine, Guava, memory-mapped files, or distributed caches like Alluxio) based on performance testing and production requirements. Similarly, the split-to-worker affinity mechanism to ensure cache locality also need to be evaluated, considering options like consistent hashing, custom node selectors, or accepting cache misses as acceptable overhead.

RFC-0022-jvector-integration.md
bibith4: Thanks for suggesting this. Updated the RFC to include the index-creation procedure call

RFC-0022-jvector-integration.md
bibith4: Thanks for suggesting this. Updated the RFC with detailed information about the ApproxNearestNeighborsFunction method and its return type.

NivinCS:
> @NivinCS : Had another question about security related features as well ? Most of the Vector search libraries used for RAG also mention about how they provide secure access. Is that a concern at all in this work ?

We had a discussion with @gopikris, and the conclusion was that this is similar to how database indexes work today. If a user has access to a table, they automatically have access to the associated vector index as well. There is no need for any separate or explicit access control for vector index.

RFC-0022-jvector-integration.md
Nandakumar-Balagopal: Thanks for sharing the pgai vectorizer doc. We did some research on the vectorizer after going through it.

Based on that, the approach we’re converging on is:
- Treat embeddings as **derived data**, not user-managed columns.
- Use a **vectorizer-style component** that maintains embeddings **asynchronously**.

In environments where a distributed processing engine is available, this can be implemented using background Spark jobs:
- Spark runs in the background to generate and refresh embeddings from the source table.
- Embeddings are stored in a **separate, system-managed table/index**.
- Updates and deletes are handled via batch or incremental Spark jobs, keeping embeddings in sync without changing the user table.

This model is **only feasible in watsonx.data**, where we have a managed Spark runtime and job orchestration.

In **OSS Presto**, this part is intentionally out of scope:
- Users would need to run their **own ingestion / embedding pipeline** (Spark, Flink, etc.).
- Presto would only support **retrieval and similarity search** over precomputed embeddings, not embedding generation or lifecycle management.

RFC-0022-jvector-integration.md
NivinCS: We had a discussion with Gopi on this, and the conclusion is that Iceberg is our current priority. While this approach can be extended to support other table formats in the future, that can be considered later. We can also look into separating the generic components from the Iceberg-specific pieces and work with Guy on the Iceberg-related aspects.

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1313 @@
...
+    index_name => 'embedding_idx',
+    similarity_function => 'COSINE',  -- COSINE, EUCLIDEAN, DOT_PRODUCT (optional)
+    m => 16,                          -- HNSW M parameter (default: 16) (optional)
+    ef_construction => 100            -- HNSW ef_construction (default: 100) (optional)
+);
+
+-- Find top 10 most similar documents to a query
+SELECT d.doc_id, d.title, ann.score
+FROM documents d
+JOIN TABLE(approx_nearest_neighbors(
```
gggrace14: It seems only one index is possibly created for an embedding column, given there's no reference to the index name, etc. And once an index is created, the supported similarity_function "COSINE" is fixed. Is it too limited in practical user scenarios?

RFC-0022-jvector-integration.md
```diff
@@ -0,0 +1,1313 @@
...
+    
+    // JSON serialization methods
+}
+
+/**
+ * Fragment data written by workers
+ * Serialized as Slice and collected by coordinator
+ */
+public class BuildIndexFragment {
+    private final String partitionKey;
```
gggrace14: Can you give an example of partitionKey? Curious if you have a way to ensure even partitioning?

PR #56, opened 2026-02-10
Add RFC for Vector Search

gggrace14:
This RFC proposes the syntax and SPI to natively support Approximate Nearest Neighbor (ANN) vector search within Presto.

### DIFF
--- a/RFC-0023-vector-search.md
+++ b/RFC-0023-vector-search.md
@@ -1,0 +1,433 @@
+# RFC-0023 for Presto: Syntax and SPI for Presto ANN Vector Search
+
+Proposers
+- Zhichen Xu
+- Ge Gao
+- Ke Wang
+- Feilong Liu
+- Amit Dutta
+- Sreeni Viswanadha
+- James Gill
+
+## Summary
+
+This RFC proposes the syntax and SPI to natively support Approximate Nearest Neighbor (ANN) vector search within Presto.
+
+## Background
+
+Vector similarity search is fundamental to AI/ML applications (RAG, recommendations, anomaly detection). Today, Presto users must either:
+- Use external vector databases (data duplication, sync overhead)
+- Use brute-force `CROSS JOIN` + distance (O(nxm), prohibitive at scale)
+
+Native ANN support enables single-system analytics+vector workloads with transparent optimization.
+
+## User Journeys
+
+##### Journey 1: Create index for specific partitions now
+
+"I want to create a vector index for ds='2026-01-01' through '2026-01-03'."
+
+```sql
+CREATE VECTOR INDEX vector_index ON TABLE candidates_table(id, embedding) WITH (
+    index_type = 'ivf_rabitq4',
+    distance_metric = 'cosine',
+    index_options = 'nlist=100000,nb=8',
+    partitioned_by = ARRAY['ds']
+) UPDATING FOR
+    ds BETWEEN '2026-01-01' AND '2026-01-03'
+```
+
+Behavior: Indices are built immediately for partitions ds=2026-01-01, 2026-01-02 and 2026-01-03.
+
+##### Journey 2: Create index metadata without populating any partition
+
+"I need the index definition created now, but the index should only build when I instruct."
+
+```sql
+CREATE VECTOR INDEX vector_index ON TABLE candidates_table(id, embedding) WITH (
+    index_type = 'ivf_rabitq4',
+    distance_metric = 'cosine',
+    index_options = 'nlist=100000,nb=8',
+    partitioned_by = ARRAY['ds']
+)
+```
+
+##### Journey 3: Populate vector index for specific partitions now
+
+"I need to populate the vector index for 2026-01-04."
+
+```sql
+UPDATE VECTOR INDEX vector_index WHERE
+    ds = '2026-01-04'
+```
+
+##### Journey 4: Search with implicit index lookup
+
+"I want to run ANN search automatically using the index registered on my table."
+
+```sql
+SELECT
+    queries.id,
+    VECTOR_SEARCH(
+        candidates.id,
+        candidates.embedding,
+        queries.embedding,
+        'cosine',
+        10
+    )
+FROM candidates CROSS JOIN queries
+WHERE
+    candidates.ds BETWEEN '2026-01-01' AND '2026-01-04'
+    AND queries.ds = '2026-01-04'
+GROUP BY
+    queries.id
+```
+
+Behavior: Optimizer uses the registered index automatically.
+
+
+## Proposed Changes
+
+### Create a Vector Index
+
+#### Create Partitioned Index
+
+Partitioned index builds a standalone index for each base table partition. Most of the prod base tables are anticipated to fall in this user scenario, where:
+- The individual table partition is large.
+- A new partition is added every day or every hour.
+- Past partitions are occasionally backfilled.
+
+```sql
+CREATE VECTOR INDEX vector_index ON TABLE candidates_table(id, embedding) WITH (
+    index_type = 'ivf_rabitq4',
+    distance_metric = 'cosine',
+    index_options = 'nlist=100000,nb=8',
+    partitioned_by = ARRAY['ds']
+) UPDATING FOR
+    ds BETWEEN '2026-01-01' AND '2026-01-03'
+```
+
+**ON `TABLE candidates_table(id, embedding)` clause**
+- `embedding` is the embedding column to create the index on.
+- `id` is the unique identifier column that the user needs to provide.
+    - Presto will not enforce the uniqueness and will leave it to the user.
+    - Alternatively, use the hidden column $row_id for id and remove the requirement from users.
+
+**`UPDATING FOR` clause**
+- For each of the index partitions mapped to the range, an index will be created.
+    - For the mapping of index partition to base table partition, refer to the table property partitioned_by.
+
+**Table properties**
+- `partitioned_by`: Partition keys of the index.
+    - For each of the index partitions, an index will be created.
+    - Must be a subset of the base table partition keys.
+- `index_type`: Index type. Different ConnectorFactory could define a different enum of index types to support.
+- `distance_metric`: Distance metric, ex. cosine similarity, inner product, L2 distance.
+- `index_options`: Options specific to the index type, in varchar type that is decoded to Map afterwards.
+
+It is allowed to omit the `UPDATING FOR` clause as below. In this case, we will only create the vector index (metadata) but not populate the per-partition indices. Users use UPDATE VECTOR INDEX below to populate the per-partition indices.
+
+```sql
+CREATE VECTOR INDEX vector_index ON candidates_table(id, embedding) WITH (
+    index_type = 'ivf_rabitq4',
+    distance_metric = 'cosine',
+    index_options = 'nlist=100000,nb=8',
+    partitioned_by = ARRAY['ds']
+)
+```
+
+#### Create Unpartitioned Index
+
+Unpartitioned index builds a single index for the base table over the partitions specified by UPDATING FOR clause. It is used for the scenario when:
+- The base table is small enough, given the range of partitions specified by UPDATING FOR clause.
+- The required latency and resource for populating the single index is acceptable to the user.
+- The user is interested in an ad-hoc index creation and ANN search.
+
+`UPDATING FOR` clause is used as below to specify the range of partition data to build the single index over.
+
+```sql
+CREATE VECTOR INDEX vector_index ON candidates_table(id, embedding) WITH (
+    index_type = 'ivf_rabitq4',
+    distance_metric = 'cosine',
+    index_options = 'nlist=100000,nb=8'
+) UPDATING FOR
+    ds BETWEEN '2026-01-01' AND '2026-01-03'
+```
+
+When no `UPDATING FOR` clause is specified, we will only create the vector index (metadata) but not populate the single index.
+
+```sql
+CREATE VECTOR INDEX vector_index ON candidates_table(id, embedding) WITH (
+    index_type = 'ivf_rabitq4',
+    distance_metric = 'cosine',
+    index_options = 'nlist=100000,nb=8'
+)
+```
+
+#### Execution and SPI
+
+It will be a general and flexible interface if we make the execution similar to CTAS with a function `create_vector_index()` introduced. Plugins could have their own implementation of the *create_vector_index()* function.
+
+- Add an AST `CreateVectorIndex` node that extends `Statement`.
+
+```java
+public class CreateTableAsSelect extends Statement { }
+```
+
+- Add a branch to `LogicalPlanner.planStatementWithoutOutput()` for `CreateVectorIndex` AST node and create the logical plan nodes.
+
+```
+TableFinishNode
+    |__ TableWriteNode
+            |__ RelationPlan (create_vector_index())
+```
+
+- Add `CreateVectorIndexReference` that extends `TableWriterNode.WriterTarget`
+
+```java
+public static class CreateVectorIndexReference extends WriterTarget { }
+```
+
+Logically it is equivalent to:
+
+```sql
+CREATE TABLE vector_index WITH (
+    partitioned_by = ARRAY['ds']
+) AS SELECT
+    create_vector_index(
+        id,
+        embedding,
+        'ivf_rabitq4',      -- index_type
+        'cosine',           -- distance_metric
+        'nlist=100000,nb=8' -- index_options
+    )
+FROM candidates_table
+WHERE
+    ds BETWEEN '2026-01-01' AND '2026-01-03'
+```
+
+#### Metadata
+
+- Introduce to `MetadataManager`
+    - `createVectorIndex()`
+    - `beginCreateVectorIndex()`
+    - `finishCreateVectorIndex()`
+- Introduce to ConnectorMetadata
+    - `createVectorIndex()`
+    - `beginCreateVectorIndex()`
+    - `finishCreateVectorIndex()`
+- HiveMetadata implementation
+    - Similar to CTAS, plus a boolean table parameter `is_vector_index`.
+
+### Update Vector Index
+
+PrestoSQL interface to rebuild the index. After the index is created and has metadata stored in Metastore, UPDATE VECTOR INDEX could be:
+- Manually triggered by the user.
+
+#### Update Partitioned Index
+
+```sql
+UPDATE VECTOR INDEX vector_index WHERE
+    ds = '2026-01-04'
+```
+
+Update only the index partition mapped to the base table partition ds = '2026-01-04'.
+
+```sql
+UPDATE VECTOR INDEX vector_index
+```
+
+Update all the indices corresponding to all the base table partitions.
+
+#### Update Unpartitioned Index
+
+```sql
+UPDATE VECTOR INDEX vector_index WHERE
+    ds = '2026-01-04'
+```
+
+Rebuild the single unpartitioned index based on the WHERE clause of the original vector index definition stored in Metadata.
+
+
+### Vector Search
+
+#### Search with a new function VECTOR_SEARCH()
+
+```sql
+SET SESSION vector_search_index = 'di:vector_index:nprobe=50';
+
+SELECT
+    queries.id,
+    VECTOR_SEARCH(
+        candidates.id,
+        candidates.embedding,
+        queries.embedding,
+        'cosine',
+        10
+    )
+FROM candidates CROSS JOIN queries
+WHERE
+    candidates.ds BETWEEN '2026-01-01' AND '2026-01-04'
+    AND queries.ds = '2026-01-04'
+GROUP BY
+    queries.id
+```
+
+#### OSS Implementation
+
+OSS implementation will be exact vector search and implement a ConnectorPlanOptimizer to rewrite to a plan with MAX_BY().
+
+```sql
+SELECT
+    queries.id,
+    MAX_BY(
+        candidates.id,
+        COSINE_SIMILARITY(candidates.embedding, queries.embedding),
+        10
+    )
+FROM candidates CROSS JOIN queries
+WHERE
+    candidates.ds BETWEEN '2026-01-01' AND '2026-01-04'
+    AND queries.ds = '2026-01-01'
+GROUP BY
+    queries.id
+```
+
+#### Metadata
+
+Rewrite needs the following metadata
+
+`vector_index` The vector index for search and its schema.
+- Introduce metastore.getVectorIndex() API.
+    - Functionality-wise, similar to metastore.getTable().
+
+### Table Function
+
+Table function provides an alternative syntax which can be more intuitive for simple use cases.
+
+```sql
+SELECT * FROM TABLE(vector_search(
+    DOCUMENTS => TABLE(candidates_table),
+    DOC_VECTOR_COL => 'embedding',
+    QUERIES => TABLE(queries_table),
+    QUERY_VECTOR_COL => 'query_embedding',
+    TOP_K => 10,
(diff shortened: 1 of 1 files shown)


NivinCS:
Thanks, @gggrace14 , for the write-up 

The proposal currently does not describe the lifecycle management of vector indexes with respect to underlying table mutations. In a lakehouse environment such as Iceberg, operations like appends, partition backfills, deletes, file rewrites (compaction), and snapshot expiration can make the index stale or physically invalid, especially if the index stores file-level row references.

Could we clarify how index validity is expected to be maintained across such operations? Specifically:

1. How is index staleness detected?
2. Are indexes expected to be snapshot-bound?
3. How should index cleanup or rebuild be handled after file rewrites or snapshot expiry?
4. Is automatic invalidation or maintenance of index metadata in scope?

This would help ensure correctness and avoid cases where ANN search may return outdated or invalid results.

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+```
+
+Behavior: Indices are built immediately for partitions ds=2026-01-01, 2026-01-02 and 2026-01-03.
+
+##### Journey 2: Create index metadata without populating any partition
+
+"I need the index definition created now, but the index should only build when I instruct."
+
+```sql
+CREATE VECTOR INDEX vector_index ON TABLE candidates_table(id, embedding) WITH (
```
NivinCS, line 47: After creating the vector index metadata without building it for any partitions, the expected query behavior is not entirely clear. For example, if a user runs:
```
SELECT *
FROM candidates_table
ORDER BY embedding <-> query_vector
LIMIT 10;
```
when the index is defined but not yet populated, should the engine fall back to a full scan, ignore the index, emit a warning, or fail the query? It would be helpful to clarify the intended behavior in this scenario to ensure predictable query execution.

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+    distance_metric = 'cosine',
+    index_options = 'nlist=100000,nb=8',
+    partitioned_by = ARRAY['ds']
+) UPDATING FOR
+    ds BETWEEN '2026-01-01' AND '2026-01-03'
+```
+
+**ON `TABLE candidates_table(id, embedding)` clause**
+- `embedding` is the embedding column to create the index on.
+- `id` is the unique identifier column that the user needs to provide.
```
NivinCS, line 112: It would be beneficial to keep the PRIMARY KEY / ROW_ID specification optional rather than mandatory for vector index creation.

From the Iceberg V3 perspective, tables with Row Level Lineage enabled can provide a stable row-level identifier via $row_id. In such cases, the identifier column can be made optional, allowing the connector to automatically fall back to $row_id when a user-defined id is not specified.

This would avoid requiring users to explicitly provide a unique identifier for tables where lineage-based row identity is already available.

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+    - Presto will not enforce the uniqueness and will leave it to the user.
+    - Alternatively, use the hidden column $row_id for id and remove the requirement from users.
+
+**`UPDATING FOR` clause**
+- For each of the index partitions mapped to the range, an index will be created.
```
NivinCS, line 117: It may be beneficial to make the partition specification in UPDATE VECTOR INDEX (e.g., FOR PARTITION (...)) optional rather than mandatory.

Instead of requiring users to manually specify partition ranges for index updates, the connector could automatically detect changed partitions and perform incremental index refresh, similar to materialized view incremental refresh.

This would reduce the need for manual orchestration and improve usability for incremental index maintenance.
cc : @tdcmeehan

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+WHERE
+    candidates.ds BETWEEN '2026-01-01' AND '2026-01-04'
+    AND queries.ds = '2026-01-04'
+GROUP BY
+    queries.id
+```
+
+#### OSS Implementation
+
+OSS implementation will be exact vector search and implement a ConnectorPlanOptimizer to rewrite to a plan with MAX_BY().
```
NivinCS, line 278: It will be beneficial to include a reusable toolkit/library as part of the proposal to handle internal query rewrites for vector search operations.

In many cases, executing ANN search would otherwise require users to explicitly join the base table with the corresponding vector index using a row identifier (e.g., user-defined id or $row_id). This can expose index implementation details and make query syntax more complex, requiring users to understand how embeddings and index data are related internally.

Providing a reusable component that transparently rewrites vector search predicates (e.g., ANN_SEARCH(...)) into the appropriate join between the base table and vector index would enable a simpler and more seamless user experience. This would allow users to express ANN queries without manually specifying join logic, while the underlying connector leverages the index automatically.

Such a library could be integrated by connectors (e.g., Iceberg) to perform these rewrites behind the scenes, avoiding duplication of rewrite logic across connectors and keeping index implementation details abstracted from end users.
cc : @tdcmeehan

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+
+##### Journey 1: Create index for specific partitions now
+
+"I want to create a vector index for ds='2026-01-01' through '2026-01-03'."
+
+```sql
+CREATE VECTOR INDEX vector_index ON TABLE candidates_table(id, embedding) WITH (
+    index_type = 'ivf_rabitq4',
+    distance_metric = 'cosine',
+    index_options = 'nlist=100000,nb=8',
```
Dilli-Babu-Godari, line 34: what do the parameters `nlist `and `nb ` represent?

1. How do these parameters affect indexing performance, recall, and query latency?
2. What guidelines should users follow to choose optimal values for different dataset sizes and vector dimensions?
3. Is there any automatic tuning mechanism available, or are users expected to benchmark and manually determine the best configuration?

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+"I need to populate the vector index for 2026-01-04."
+
+```sql
+UPDATE VECTOR INDEX vector_index WHERE
+    ds = '2026-01-04'
+```
+
+##### Journey 4: Search with implicit index lookup
+
+"I want to run ANN search automatically using the index registered on my table."
```
aditi-pandit, line 66: Does the optimizer know about the Index as such ? The use of the index should be abstracted by VECTOR_SEARCH function, no ?

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+    - Similar to CTAS, plus a boolean table parameter `is_vector_index`.
+
+### Update Vector Index
+
+PrestoSQL interface to rebuild the index. After the index is created and has metadata stored in Metastore, UPDATE VECTOR INDEX could be:
+- Manually triggered by the user.
+
+#### Update Partitioned Index
+
+```sql
```
aditi-pandit, line 229: Are we missing the SPI for these ? Can you elaborate ?

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+
+
+### Vector Search
+
+#### Search with a new function VECTOR_SEARCH()
+
+```sql
+SET SESSION vector_search_index = 'di:vector_index:nprobe=50';
+
+SELECT
```
aditi-pandit, line 259: Have you considered implementing VECTOR_SEARCH with a TABLE function ? It is easier to express the input tables, partitioning and parameters with it.

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+    TOP_K => 10,
+    DOC_ID => 'id',
+    QUERY_ID => 'id',
+    DISTANCE_METRIC => 'cosine'
+))
+```
+
+#### Execution and SPI
+
+Our proposal is to introduce generic SPI nodes: VectorSearchNode and VectorSearchOptions, which together provide an abstract interface for vector search. This abstract interface will only include mandatory parameters, allowing actual implementations to extend it.
```
aditi-pandit, line 323: Can you give more details about VectorSearchNode and VectorSearchOptions. Will VectorSearchNode be a PlanNode ?

aditi-pandit (commented):
Thanks @gggrace14 for this writeup. Overall I'm in favor of this proposal.

I'm also curious since FAISS functions were recently exposed in C++.... Are you planning to do any follow up work with the library ?

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+
+### Create a Vector Index
+
+#### Create Partitioned Index
+
+Partitioned index builds a standalone index for each base table partition. Most of the prod base tables are anticipated to fall in this user scenario, where:
+- The individual table partition is large.
+- A new partition is added every day or every hour.
+- Past partitions are occasionally backfilled.
+
```
bibith4, line 99: It would be helpful to clarify the behavior when base table partitions are dropped, since index partitions are mapped to them. Should corresponding index partitions be automatically cleaned up or require manual action?

Additionally, it may be beneficial to define lifecycle DDL such as DROP VECTOR INDEX for operational usability.

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+```sql
+UPDATE VECTOR INDEX vector_index WHERE
+    ds = '2026-01-04'
+```
+
+Update only the index partition mapped to the base table partition ds = '2026-01-04'.
+
+```sql
+UPDATE VECTOR INDEX vector_index
```
bibith4, line 237: It would be helpful to clarify query behavior during index build or refresh. For example, can queries read partially built partitions?

gggrace14:
> Thanks, @gggrace14 , for the write-up
> 
> The proposal currently does not describe the lifecycle management of vector indexes with respect to underlying table mutations. In a lakehouse environment such as Iceberg, operations like appends, partition backfills, deletes, file rewrites (compaction), and snapshot expiration can make the index stale or physically invalid, especially if the index stores file-level row references.
> 
> Could we clarify how index validity is expected to be maintained across such operations? Specifically:
> 
> 1. How is index staleness detected?
> 2. Are indexes expected to be snapshot-bound?
> 3. How should index cleanup or rebuild be handled after file rewrites or snapshot expiry?
> 4. Is automatic invalidation or maintenance of index metadata in scope?
> 
> This would help ensure correctness and avoid cases where ANN search may return outdated or invalid results.

Hi @NivinCS , in general it is designed to define the minimum required parts to be SPI and leave individual connectors with the max flexibility to define its own behavior.

Specifically with respect to how index validity is maintained, we could leave it to individual ConnectorPlanOptimizer to detect if the index partition is stale and return reasonable result when the ConnectorPlanOptimizer plans/rewrites the vector search query.

For SPI, thinking through your questions, I think what could be added here is a `getVectorSearchIndexStatus()` method in the SPI class MetadataManager and ConnectorMetadata, similar to [how materialized view detects staleness](https://github.com/prestodb/presto/blob/67304707b442bc120e3402ab270b8e59135755e8/presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorMetadata.java#L776). Let me know if there are something else in your mind that need to be added to SPI.

I could describe below what we will put into the HiveConnector to maintain the index partition validity at Meta, by answering your above questions. Meta has an internal HiveConnector extension. I think you can implement the Iceberg behaviors in the counterpart classes of IcebergConnector. Again I think these should belong to the connectors but not SPI.

**1. How is index staleness detected?**
By looking at the metadata of index partition on HiveMetastore. As index partition is just a Hive table partition, if the index partition is stale, we'll drop the partition from the HiveMetastore. We can also put a dirty mark on the metadata of the index partition.

We'll have a pub-sub service that subscribes to the metadata of the candidate table (base table). If the corresponding base table partition is updated/overwritten/deleted, it will trigger the dropping of the index partition.

Then in the search path, we have a VectorSearchRewritePlanOptimizer that implements a ConnectorPlanOptimizer. In the VectorSearchRewritePlanOptimizer, we compare the available partitions of the index vs the base table in the query range. If an index partition is available, we'll use it to get the kNN of the partition. Otherwise, we will build the index on the fly. Another option for an unavailable index partition is to fall back to the exact search. And then we'll aggregate across all partitions in the query range and compute the kNN from all kNNs from each partion.

**2. Are indexes expected to be snapshot-bound?**
I might not have enough context for Iceberg. However, I think you can decide the reasonable behavior in IcebergConnector.

**3. How should index cleanup or rebuild be handled after file rewrites or snapshot expiry?**
As mentioned above, we will have a pub-sub service. Once it detects the update or deletion event of the base table from HiveMetastore, it will trigger at least two actions in sequence, 1) drop the corresponding index partition; 2) call `UPDATE VECTOR INDEX vector_index WHERE <filter>`, where <filter> corresponds to the dropped partition.

**4. Is automatic invalidation or maintenance of index metadata in scope?**
No, the automatic invalidation is out of the scope of Presto as well as this RFC. The index validity is stored as the index metadata. And when and how to update the index metadata is out of the scope of Presto. Presto only provides the SQL as interface to create and update the index.

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+    partitioned_by = ARRAY['ds']
+)
+```
+
+##### Journey 3: Populate vector index for specific partitions now
+
+"I need to populate the vector index for 2026-01-04."
+
+```sql
+UPDATE VECTOR INDEX vector_index WHERE
```
jja725, line 60: UPDATE VECTOR INDEX: This looks like a DML UPDATE but is actually a rebuild/refresh operation. REFRESH VECTOR INDEX or REBUILD VECTOR INDEX would be clearer and avoid confusion with
  standard SQL UPDATE semantics.

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+```
+
+#### Execution and SPI
+
+It will be a general and flexible interface if we make the execution similar to CTAS with a function `create_vector_index()` introduced. Plugins could have their own implementation of the *create_vector_index()* function.
+
+- Add an AST `CreateVectorIndex` node that extends `Statement`.
+
+```java
+public class CreateTableAsSelect extends Statement { }
```
jja725, line 174: should be CreateVectorIndex?

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+UPDATE VECTOR INDEX vector_index
+```
+
+Update all the indices corresponding to all the base table partitions.
+
+#### Update Unpartitioned Index
+
+```sql
+UPDATE VECTOR INDEX vector_index WHERE
+    ds = '2026-01-04'
```
jja725, line 246: This is confusing — does the WHERE clause here replace the original filter, extend it, or is it ignored? Why we need this filter if it's unpartitioned

RFC-0023-vector-search.md
gggrace14: Hi @Dilli-Babu-Godari , generally index_options is particular to individual index_type. And each connector is responsible for checking if the index_type is supported and if the index_options is valid for the index_type. (I'm adding this detail to this RFC)

Specifically,
_1. How do these parameters affect indexing performance, recall, and query latency?_

`nlist` is one option for any ivf_* index, which is the # of centroids. The typical trade-off around the # of centroids also hold here. Larger `nlist` means smaller cluster size. If the data distribution itself is good, it could largely reduce the search space, and thus reduce the amount of records loaded to each node as well as memory pressure. If the data distribution is not good, then we need to increase `nprobe`.

`nb` is one option for the index RabitQ. Higher nb provides better accuracy at the cost of additional storage, as in https://github.com/facebookresearch/faiss/wiki/Guidelines-to-choose-an-index#if-maximum-compression-is-required-then-rabitq.

_2. What guidelines should users follow to choose optimal values for different dataset sizes and vector dimensions?_
For `nlist`, we can use the formula here, https://github.com/facebookresearch/faiss/wiki/Guidelines-to-choose-an-index#if-below-1m-vectors-ivfk.
For `nb`, more bits means better accuracy but at the cost of memory.

_3. Is there any automatic tuning mechanism available, or are users expected to benchmark and manually determine the best configuration?_
At the moment, we derive default values in code for the options, if possible. This index_options parameter is what the user can use to customize the options. Automatic tuning is next next step, which needs the index quality feedback loop.

RFC-0023-vector-search.md
```diff
@@ -0,0 +1,433 @@
...
+## Design Rationale
+
+1. **BigQuery-aligned syntax**  
+   CREATE VECTOR INDEX and VECTOR_SEARCH table function provide familiar semantics
+
+2. **Partition-aware indexing**  
+   Unique to Presto; critical for large-scale warehouse workloads with daily partitions
+
+3. **Pre-filtering advantage**  
+   WHERE before search enables partition pruning (BigQuery lacks this)
```
jja725, line 427: https://docs.cloud.google.com/bigquery/docs/vector-index#pre-filters_and_post-filters bigquery actually have pre-filter seems like

jja725 (commented):
Hi @gggrace14 , thanks for the proposal and great to see you again in the community! This definitely would help presto involve in the AI/ML workloads. I recently worked on lance connector and and would be happy to jump in and help with the vector search integration!

RFC-0023-vector-search.md
gggrace14: It is by current design for the RFC to leave it to individual connector defining the behavior, when the index only has schema but has no data.

For example, at meta internal the current strategy is to have a condition, and build the index on the fly if the condition is met, and fall back to full scan otherwise. And we put this strategy into a ConnectorPlanOptimizer subclass.

I think it's too early to define such behavior or any behavior in SPI, and let every connector respect this behavior. A better approach is to let individual connector contribute the implementation first, and we extract the common behavior into SPI afterwards.

RFC-0023-vector-search.md
gggrace14: Hi @NivinCS , sure, make sense. I will change rfc to make the `id` field optional.

RFC-0023-vector-search.md
gggrace14: By design `UPDATE VECTOR INDEX vector_index WHERE <filter>` should be the Presto interface for incremental index refresh, similar to the `REFRESH MATERIALIZED VIEW ... WHERE <filter>`. I think the auto detection of partition change to the base table should be out of the scope of Presto. For example, we could possibly have a pub-sub service out of Presto that auto detects based on the base table metadata. Once the change to base table partition is detected, any auto mechanism could assemble the <filter> and call `UPDATE VECTOR INDEX` to instruct Presto to refresh the index increment.

RFC-0023-vector-search.md
gggrace14: Whether the candidates table has an index or not would be maintained by metadata, which is accessible to optimizer. If the index is available, the optimizer could pick it up and rewrite the VECTOR_SEARCH query.

RFC-0023-vector-search.md
gggrace14: Hi @aditi-pandit , yes, we do plan to implement VECTOR_SEARCH as a TABLE function, on top of the table function framework.

Yea, also thought to follow up. In the prior TSC meeting, you mentioned that you have some experience or resources that you could share with us to implement the VECTOR_SEARCH function as a table function. Would you kindly provide more details? Thank you!

RFC-0023-vector-search.md
gggrace14: Sorry, Aditi, I think this paragraph is outdated. I'm removing this paragraph.

RFC-0023-vector-search.md
gggrace14: Yea, I'm adding the DDL `DROP VECTOR INDEX`. A good suggestion.

I'm changing this RFC to add a `getVectorSearchIndexStatus()` method to the SPI class MetadataManager and ConnectorMetadata. Then the search path could decide whether to consume the index partitions given the index status.

Regarding the behavior when the base table partitions are dropped, I think we should leave to users to decide whether to drop the index partitions or not. In some case, the index building is quite expensive, and the user might not want to drop the index partitions until the index quality does not meet certain criteria. We could make suggestion in the RFC. At meta, currently we drop the index partitions. I'm also adding this clarification to the RFC.

RFC-0023-vector-search.md
gggrace14: We can add a `getVectorSearchIndexStatus()` method to the SPI class MetadataManager and ConnectorMetadata. It's better to leave it to the individual ConnectorPlanOptimizer to decide whether to pick up the partially built index partitions or now.

RFC-0023-vector-search.md
gggrace14: Hi @jja725 , I'm open with REFRESH VECTOR INDEX or REBUILD VECTOR INDEX, if no one else objects to. I could change this RFC to use REFRESH VECTOR INDEX.

RFC-0023-vector-search.md
gggrace14: You're right. Correcting the type now.

RFC-0023-vector-search.md
gggrace14: I think the confusing part is that `UPDATE VECTOR INDEX` sounds like an update to the index schema. Here the semantics is actually to refresh the index partition `ds = '2026-01-04'`. I can change the `UPDATE VECTOR INDEX` to `REFRESH VECTOR INDEX`, same to the reply to your previous comment. By design, we don't support mutating the index schema, or `replace vector index`.

gggrace14:
> Thanks @gggrace14 for this writeup. Overall I'm in favor of this proposal.
> 
> I'm also curious since FAISS functions were recently exposed in C++.... Are you planning to do any follow up work with the library ?

Hi @aditi-pandit , thank you for reviewing this RFC in the TSC meeting and here. Yes, at meta so far we've accumulated implementation for the HiveConnector on top of the FAISS library, for both the search path and `create vector index` path. And we have plan to move some implementation and, more importantly, the algorithm to open source as common lib or toolkit. The challenge is we're building a sophisticated two-level hierarchical search algorithm in order to scale to >1 billion internal dataset. We need to figure out a clever way to extract components that are universally reusable for the presto community. We've also written a paper describing this approach. I think the paper description is the best way to share the algorithm first, once accepted :)

RFC-0023-vector-search.md
jja725: REFRESH VECTOR INDEX make sense.

NivinCS: I agree with the suggestion to use REFRESH VECTOR INDEX instead of UPDATE VECTOR INDEX.

If the operation is not performing row-level modifications like a standard SQL UPDATE, but instead rebuilding or synchronizing the index with the latest table state, using REFRESH makes the intent clearer and avoids confusion with DML semantics.

That said, I have one question for clarification: are we expecting to rebuild the entire index each time? For large datasets (e.g., billions of records), a full rebuild could be quite expensive. Would it make sense to also consider support for incremental updates, as many modern vector search systems allow updating the index on the fly?

gggrace14: Hi @NivinCS , thank you for your inputs! Will use REFRESH VECTOR INDEX.

Regarding incremental building of an index, I can share our thoughts and implementation so far at meta as below
1) As you know we look at Hive tables at the moment. So we partition the index and build an index per base table partition. This way allows us to only rebuild the index for one partition each time. Merging of per-partition results is done at the search path.

2) We actually have two levels of indices, and only building the first level is expensive. The first level index could be reused and don't have to refresh every time.

3) Underlying we use Faiss index. And faiss has some effort exploring incremental index building.

Hope it helps. In general, we put incremental index building into the concrete implementation of REFRESH VECTOR INDEX.

NivinCS: Thanks for the clarification Ge.

NivinCS (approved):
Thanks for the proposal.

jja725 (approved):
Thanks for the contribution!

PR #59, opened 2026-04-14
Add RFC to introduce Bolt backend for native engine

Weixin-Xu:
## Summary

Introduce [Bolt](https://github.com/bytedance/bolt) as an additional backend for the Presto native execution engine.

The initial implementation provides a Bolt-based native worker that implements the Presto worker protocol and integrates with the existing Presto coordinator.

To support the Bolt backend build and dependency requirements, a Conan-based dependency flow is introduced for this worker module. Standardizing dependency management across all native backends is out of scope for this RFC.

@frankobe @ZacBlanco


beinan:
Looking forward to having bolt in native presto workers!

RFC-0024-bolt-backend.md
```diff
@@ -0,0 +1,356 @@
...
+At a high level, the Bolt worker follows this flow:
+
+1. The Java coordinator creates the same Presto plan fragments it already produces today.
+2. The Bolt worker deserializes those fragments using the existing Presto protocol model.
+3. Bolt-specific converters translate Presto plans, expressions, splits, and connector handles into Bolt-native structures.
+4. Bolt executes the resulting plan and reports task state, results, and metrics back through the existing worker protocol.
+
+The current code includes dedicated Bolt converters such as:
+
+* `PrestoToBoltQueryPlan`
```
jaystarshot, line 96: > 2. The Bolt worker deserializes those fragments using the existing Presto protocol model

How will future divergence in the protocol with Velox be handled? What happens if Velox requires a protocol change that isn’t compatible with Bolt, and vice versa?

Weixin-Xu: It depends on where the change originates.
> If it’s a change to the Presto protocol, then it becomes a contract change, and we need to ensure both Bolt and Velox can handle it correctly (ideally in a backward-compatible way).
> If it’s a Bolt- or Velox-specific interface change, then it should be handled within their respective translation/converter layers.

RFC-0024-bolt-backend.md
```diff
@@ -0,0 +1,356 @@
...
+* `PrestoToBoltExpr`
+* `PrestoToBoltConnector`
+* `PrestoToBoltSplit`
+* `BoltPlanConversion` and `BoltPlanValidator`
```
aditi-pandit, line 100: Does all the side-car code...the callbacks here : https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L1851 remain the same ?

RFC-0024-bolt-backend.md
```diff
@@ -0,0 +1,356 @@
...
+* same small container smoke topology
+* same Docker build profile shape for coordinator + worker images
+
+The main thing that changes between backends is the worker binary or worker image. The coordinator artifact, test query patterns, and data conventions stay the same.
+
+Short term, the code keeps Bolt tests in a separate module so backend-specific jobs remain isolated. Long term, we can reduce duplicated Java test sources further, but the immediate objective is to avoid duplicating heavyweight runtime resources and datasets.
+
+### 5. CI Plan
+
+CI for Bolt should be split into a few clear lanes:
```
aditi-pandit, line 310: We have a fairly exhaustive sets in the presto-native-tests module https://github.com/prestodb/presto/tree/master/presto-native-tests. Please ensure these are covered as well.

RFC-0024-bolt-backend.md
```diff
@@ -0,0 +1,358 @@
...
+
+The current code includes dedicated Bolt converters such as:
+
+* `PrestoToBoltQueryPlan`
+* `PrestoToBoltExpr`
+* `PrestoToBoltConnector`
+* `PrestoToBoltSplit`
+* `BoltPlanConversion` and `BoltPlanValidator`
+
+This is intentionally backend-local. The initial implementation does not try to share plan conversion logic with the Velox backend.
```
jja725, line 102: can we have the function-coverage delta: which Velox functions are not yet in Bolt, and which Bolt functions don't match Velox semantics?

RFC-0024-bolt-backend.md
```diff
@@ -0,0 +1,358 @@
+# RFC-0023: Introduce Bolt Backend for Presto Native Execution
```
jja725: we already have RFC-0023-vector-search.md, change the number

RFC-0024-bolt-backend.md
```diff
@@ -0,0 +1,358 @@
...
+These are reused directly from the sibling `presto-native-execution` source tree rather than from a new shared library.
+
+### What Stays Backend-Local
+
+The following remain backend-specific in the current implementation:
+
+* worker server implementation
+* task execution logic
+* operators
+* plan, expression, connector, and split conversion
```
jja725, line 128: The coordinator's planner produces one plan, and the RFC names BoltPlanValidator as the Bolt-side analog of getVeloxPlanValidator(). But validation is downstream of plan emission — by the time the worker rejects a plan, the query has already been sent. It might be better to have a coordinator-side capability description so the planner can avoid emitting plans the deployed backend can't run. The RFC should at least call out that this gap exists and how it's
  bridged for the homogeneous-pool case.

RFC-0024-bolt-backend.md
Weixin-Xu: Bolt execution aims to cover all side-car callbacks and extend them where needed.

RFC-0024-bolt-backend.md
Weixin-Xu: End-to-end tests will use the same test module wherever possible.

RFC-0024-bolt-backend.md
Weixin-Xu: Function-coverage will be reflected in the bolt/bolt-execution unittests and match Presto semantics.

RFC-0024-bolt-backend.md
```diff
@@ -0,0 +1,358 @@
...
+
+---
+
+## Summary
+
+This RFC introduces Bolt as an additional backend for Presto's native worker implementation.
+
+The initial implementation keeps the existing Velox-based worker unchanged and adds a sibling module, `presto-bolt-execution`, that implements the same Presto worker protocol against Bolt. The coordinator, query protocol, and external worker model remain unchanged.
+
+The current code does not turn `presto-native-execution` into a generic shared framework. Instead, it adds a Bolt-specific worker tree and extracts only a small set of reusable helpers from `presto-native-execution`. Build enablement is also separate in the initial implementation: Velox and Bolt are built from different module directories and produce different worker binaries from different build roots.
```
yingsu00, line 21: Where will the small set of reusable helpers reside?

RFC-0024-bolt-backend.md
```diff
@@ -0,0 +1,358 @@
...
+
+* [prestodb/rfcs#59](https://github.com/prestodb/rfcs/pull/59)
+* [Weixin-Xu/presto#1](https://github.com/Weixin-Xu/presto/pull/1)
+
+---
+
+## Summary
+
+This RFC introduces Bolt as an additional backend for Presto's native worker implementation.
+
```
yingsu00, line 18: Presto currently has other native modules at root level:
- presto-native-sidecar-plugin
- presto-native-tests
What's the plan for presto-bolt-execution to work with them? Would presto-native-tests be used to cover both Velox and Bolt?

RFC-0024-bolt-backend.md
yingsu00, line 21: Given Presto already has multiple top level native modules and it's unclear if the native side car and native tests can work with Bolt, maybe we can consider add a common top level folder to host the common helpers?

RFC-0024-bolt-backend.md
yingsu00: > > 2. The Bolt worker deserializes those fragments using the existing Presto protocol model
> 
> How will future divergence in the protocol with Velox be handled? What happens if Velox requires a protocol change that isn’t compatible with Bolt, and vice versa?

I have the same question too. It's actually not just about Presto protocol but a very general concern overall. I think the right way to handle such kind of concerns is to support **versioning** on common interfaces like the Presto SPI, the Presto communication protocol, Velox interfaces, etc. In the past people have been very cautious when changes need to be made on Presto SPI, but changes were made very frequently and freely on Velox side, like the connector interfaces, DWIO interfaces, and Presto protocol. This caused lots of rebase conflicts in our internal repo in the past. I hope Bytedance Bolt can do a better job on this in the future. 

So I see this as an opportunity to start cleaning things up, and maybe we can start working on versioning support on protocol in Presto and Bolt repos first.

RFC-0024-bolt-backend.md
Weixin-Xu: We could add a top-level presto-native-common-helper module to provide a unified abstraction layer for reusable native integration helpers.

RFC-0024-bolt-backend.md
Weixin-Xu: Before the PR is finalized, presto-bolt-execution will be validated through presto-native-tests. This should help ensure Bolt stays aligned with existing Presto behaviors and test coverage.

For presto-native-sidecar-plugin, we have not tried integrating with it yet. This will be part of our next-step investigation and integration plan.

rschlussel (commented):
I have some concerns about the overall idea. Mainly, it boils down to this adding significant complexity/overhead, and the benefit isn't clear to me. What does Presto get out of having Bolt as a backend?
Some of the challenges are:
1. as mentioned by some other reviewers, it makes any protocol changes much harder if we need to worry about two backends
2. There's a huge risk of correctness issues/behavior differences between the two backends.  The java -> Prestissimo migration had to contend with so many correctness issues, and that was meant to be a one way migration.  Even with a test suite, there just will definitely be corner cases where there will be correctness differences between the two, and we will be dealing with that risk indefinitely
3. maintenance. Once we add this support we will need to maintain it, even if the people originating it move on to other things. Is that something we want to take on as a project? Again, it comes down to that I'm not clear on what the advantages are of having this backend.

RFC-0024-bolt-backend.md
Weixin-Xu: For native sidecar: the existing sidecar integration is still Velox-shaped today. Bolt clusters will run with the sidecar plugin disabled until Bolt registers its own callbacks behind the same HTTP endpoint shapes.

For presto-native-tests: the harness itself is backend-neutral. We will reuse it as-is against the Bolt worker via PRESTO_SERVER, with a small Bolt-specific skip list for gaps relative to Presto Java, and shrink that list over time.

For the common top-level folder: agreed. I updated the RFC to call out a follow-up PR that moves shared thrift inputs, protocol base YAML, and helper headers into a new top-level presto-native-common/, so Velox and Bolt can depend on it as sibling modules.

RFC-0024-bolt-backend.md
Weixin-Xu: +1 on the goal. Frequent protocol changes causing rebase pain is exactly the kind of thing we'd like to avoid going forward. Bolt is happy to be an early adopter — we can start a versioning prototype in the Bolt repo, and contribute in the Presto repo once that work kicks off. We'll align endpoints like /v1/bolt/plan to whatever scheme lands. Probably worth a dedicated RFC, but we'd like this to move in parallel with Bolt rather than block it.

RFC-0024-bolt-backend.md
Weixin-Xu: This is a real gap in the current shape. The PR only ships worker-side validation through `BoltPlanValidator`, so an unsupported plan can fail after dispatch.

For the homogeneous-pool case, the planned bridge is to expose a static capability description on the coordinator so planning can prune unsupported features earlier, and to wire `/v1/bolt/plan` as a pre-schedule check before sending the query to Bolt workers.

I updated the RFC Risks/Limitations section to call this out explicitly. The remaining open question is broader multi-backend pool selection, where the planner would need to choose a compatible backend pool per query without introducing mixed-backend scheduling in this RFC.

PR #60, opened 2026-04-17
docs: add RFC for Query Results Cache

imjalpreet:
### Abstract

This RFC proposes a Query Results Cache for Presto that caches completed `SELECT` query results so semantically equivalent future queries skip execution entirely. The cache intercepts at `SqlQueryExecution.start()` after optimization, before scheduling. On a cache hit, pages are read from `TempStorage` and fed directly to the client via the existing `OutputBuffer → ExchangeClient → HTTP` response pipeline, bypassing scheduling and execution entirely.

### DIFF
--- a/RFC-0022-query-results-cache.md
+++ b/RFC-0022-query-results-cache.md
@@ -1,0 +1,384 @@
+# **RFC 0022: Query Results Cache**
+
+Proposers
+
+* Tim Meehan (@tdcmeehan)
+* Jalpreet Singh Nanda (@imjalpreet)
+
+## Abstract
+
+This RFC proposes a Query Results Cache for Presto that caches completed `SELECT` query results so semantically equivalent future queries skip execution entirely. The cache intercepts at `SqlQueryExecution.start()` after optimization, before scheduling. On a cache hit, pages are read from `TempStorage` and fed directly to the client via the existing `OutputBuffer → ExchangeClient → HTTP` response pipeline, bypassing scheduling and execution entirely.
+
+---
+
+## Motivation
+
+Presto users frequently re-execute identical or semantically equivalent analytical queries, dashboard refreshes, repeated report runs, BI tool polling. Each re-execution incurs the full cost of scheduling, split enumeration, and execution despite producing an identical result. A query results cache eliminates this redundant work, reducing latency for repeated queries and load on underlying data sources.
+
+The design builds on two existing Presto subsystems:
+
+- **HBO canonicalization**: provides a deterministic plan hash as cache key, including connector version metadata (e.g., Iceberg snapshot IDs) so data changes automatically produce different hashes.
+- **TempStorage**: pluggable storage SPI (local disk, S3) for serialized result pages.
+
+This follows the same pattern as the Fragment Result Cache, intercept at the execution boundary, applied at whole-query level on the coordinator instead of per-split on workers.
+
+```
+SQL → Parse → Analyze → Optimize → Compute plan hash (includes table version)
+   ├── HIT:  Read pages from TempStorage → OutputBuffer → Client
+   └── MISS: Fragment → Schedule → Execute → Capture pages → Store
+```
+
+---
+
+## Goals
+
+- Eliminate redundant execution of identical `SELECT`s.
+- Single storage layer via `TempStorage` SPI for both metadata and result pages.
+- Version-aware invalidation: cache keys include connector-specific version metadata (e.g., Iceberg snapshot IDs), so any data mutation produces a different hash and is an automatic cache miss. Only connectors that declare `SUPPORTS_RESULT_CACHE` (guaranteeing ACID-consistent handle identity) are eligible.
+
+## Non-Goals (v1)
+
+- Partial reuse / predicate stitching
+- DML caching
+- Non-deterministic function caching
+- KMS envelope encryption (deferred to v2)
+
+---
+
+## Proposal
+
+### Cache Key
+
+SHA-256 of the canonicalized query plan combined with a session properties fingerprint, computed via HBO's `CanonicalPlanGenerator`.
+
+Uses a new `RESULT_CACHE` canonicalization strategy. Like `CONNECTOR`, it supports all plan node types and delegates to `ConnectorTableLayoutHandle.getIdentifier()` for connector-specific normalization. Unlike `CONNECTOR` and the other HBO strategies, it preserves all constants, filter predicates, projection constants, and scan predicates are never stripped. Two queries with different predicate values produce different cache keys.
+
+Canonicalization normalizes variable names to positional (`_col_0`, ...), sorts predicates/join criteria, and strips source locations. Serialized to JSON with deterministic key ordering, then hashed.
+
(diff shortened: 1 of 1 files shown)


RFC-0022-query-results-cache.md
```diff
@@ -0,0 +1,384 @@
...
+
+---
+
+### Read Path
+
+In `SqlQueryExecution.start()`, after `createLogicalPlanAndOptimize()`:
+
+1. Compute canonical plan hash.
+2. Check `TempStorage.exists()` on the metadata handle for the key.
+3. On exists: read and deserialize `QueryResultsCacheEntry` from metadata file. Validate: schema match (column names + types vs current `OutputNode`), expiration check, whole-result `resultHash` check.
```
agrawalreetika, line 172: Does output column order participate in cache validation? The RFC mentions validating column names + types, but it may help to explicitly mention column ordering as well.

RFC-0022-query-results-cache.md
```diff
@@ -0,0 +1,384 @@
...
+
+`QueryResultsCacheWriter` hooks into the query lifecycle via `addFinalQueryInfoListener` (same pattern as `HistoryBasedPlanStatisticsTracker`).
+
+**Page capture**: tee-write at the `ExchangeClient`:
+
+As pages arrive at the coordinator's `ExchangeClient` during normal execution, a cache-aware listener asynchronously writes each batch to `TempStorage` in the background. The client receives pages with normal latency, cache writes are off the critical path.
+
+- A running byte counter tracks total result size. If it exceeds `max-result-size`, the tee-write is abandoned and partial files are cleaned up.
+- On successful `SELECT` completion, the `QueryResultsCacheWriter` collects the accumulated `TempStorageHandle` references, builds a `QueryResultsCacheEntry`, and stores metadata in the cache provider.
+- If the query fails or is cancelled, partial cache writes are discarded.
```
agrawalreetika, line 190: How are orphaned/partial cache objects cleaned up? Or cache write starts only after query is completely FINISHED?
For example, if page uploads succeed but metadata.json write fails, or the coordinator crashes during cache population. Is cleanup purely TTL-based?

RFC-0022-query-results-cache.md
```diff
@@ -0,0 +1,384 @@
...
+### Read Path
+
+In `SqlQueryExecution.start()`, after `createLogicalPlanAndOptimize()`:
+
+1. Compute canonical plan hash.
+2. Check `TempStorage.exists()` on the metadata handle for the key.
+3. On exists: read and deserialize `QueryResultsCacheEntry` from metadata file. Validate: schema match (column names + types vs current `OutputNode`), expiration check, whole-result `resultHash` check.
+4. On hit: `serveCachedResult()` — read `SerializedPage` objects from `TempStorage`, verify per-page CRC, enqueue into `OutputBuffer`, transition to finishing. Client sees no difference from normal execution.
+5. On miss: register query ID + cache key for population on completion, proceed with normal scheduling.
+
```
agrawalreetika, line 175: How is resource usage accounted for on cache hits? Since cached queries bypass tasks/operators/exchanges:

1. Are they still counted against Resource Groups/concurrency limits?
2. How are coordinator CPU/network/memory costs tracked?

RFC-0022-query-results-cache.md
```diff
@@ -0,0 +1,384 @@
...
+**Whole-result hash (`resultHash`)**: a CRC-32C computed over all `pageCrcs` concatenated in page order, stored in the metadata. This is checked first on read, before individual pages are fetched, as a cheap early signal that the entry is intact. A mismatch skips fetching pages entirely and falls through immediately to normal execution.
+
+CRCs are computed over the **encrypted** page bytes, not the plaintext. Integrity and confidentiality are therefore independent: corruption is detectable without decrypting the page first, and the CRC does not leak plaintext structure.
+
+Any integrity check failure (per-page or whole-result) is treated identically to a cache miss. The query proceeds to normal execution and the corrupted entry is overwritten on completion via the normal write path. Failures are emitted as a metric (`cache.integrity_failure_count`) for visibility.
```
agrawalreetika, line 160: Do we plan to expose cache-hit observability/metrics?
For example:
1. servedFromResultCache
2. EventListener visibility

RFC-0022-query-results-cache.md
```diff
@@ -0,0 +1,384 @@
...
+---
+
+### Read Path
+
+In `SqlQueryExecution.start()`, after `createLogicalPlanAndOptimize()`:
+
+1. Compute canonical plan hash.
+2. Check `TempStorage.exists()` on the metadata handle for the key.
+3. On exists: read and deserialize `QueryResultsCacheEntry` from metadata file. Validate: schema match (column names + types vs current `OutputNode`), expiration check, whole-result `resultHash` check.
+4. On hit: `serveCachedResult()` — read `SerializedPage` objects from `TempStorage`, verify per-page CRC, enqueue into `OutputBuffer`, transition to finishing. Client sees no difference from normal execution.
```
agrawalreetika, line 173: What query states will cache-hit queries go through? Since execution is bypassed, will queries still transition through RUNNING as expected by the UI/client protocol?

RFC-0022-query-results-cache.md
aaneja: On similar lines, what are the operator level metrics that will show up for such a query ?  

To me a cached-query-result is akin to doing a rewrite for this query to a MV or a (materialized) CTEProducerNode, so I would expect to see the plan + metrics related to reading data from such a proxy-source-node. Maybe we can call it a `CachedQueryResults` plan node ?

RFC-0022-query-results-cache.md
```diff
@@ -0,0 +1,384 @@
...
+| 4 | Encryption: `AesSpillCipher` integration, DEK in metadata, CRC over encrypted bytes |
+| 5 | Invalidation + cleanup: TTL, access-based extension, bypass/invalidate session properties, stale lock handling |
+| 6 | S3 `TempStorage` plugin: new Maven module, conditional writes, expiration buffer |
+| 7 | Testing: unit, integration, performance, security (RLS isolation) |
+
+---
+
+## Future Work
+
+- **Predicate stitching**: Partial cache reuse for partition-decomposable queries when only some partitions change. Requires per-partition stats.
```
aaneja, line 380: +1. I think we could extend the idea to do sub-plan matching, similar to how we're exploring MV matching

amitkdutta:
For Purging the cache, wondering if we are thinking any approach or just delete the cache directory upon restart. Also wonderign what happens if the local tempstroage is out of capacity as lots of query results will be stored there.

RFC-0022-query-results-cache.md
```diff
@@ -0,0 +1,384 @@
...
+
+Canonicalization normalizes variable names to positional (`_col_0`, ...), sorts predicates/join criteria, and strips source locations. Serialized to JSON with deterministic key ordering, then hashed.
+
+**Session properties fingerprint**: a deterministic hash of the subset of session properties that can affect query results. This includes any connector specific session properties(e.g., `legacy_timestamp`, etc.) that influence scan or execution behavior. Properties that are purely operational (e.g., `query_max_execution_time`, `query_priority`) are excluded from the fingerprint as they do not affect result content. The set of result-affecting properties is declared explicitly via a new `SessionPropertyManager.getResultAffectingProperties()` method, so additions to Presto's session property surface are opt-in to the fingerprint rather than silently excluded by default.
```
tdcmeehan: Please ensure that deterministic but session dependent functions are additionally excluded from the result cache.  For example, if a function changes depending on configuration or session information, it should be excluded lest the cache include session information that might not replicate to other sessions.

imjalpreet: @tdcmeehan, thank you for the suggestion.

Yes, I have included a separate check to detect deterministic but session-dependent functions, such as current_timestamp, etc. I’ll update the RFC to call this out explicitly as well.

Some session information, such as Session.timeZoneKey and Session.locale, is also included in the cache key.

I’ll evaluate whether there are any other session-dependent functions that need to be considered. Please let me know if you had any specific ones in mind.

PR #61, opened 2026-04-24
RFC-0025 Derived column

ScrapCodes:
What is a derived column?

A column created by applying a SQL expression or a UDF to an existing column in a table.

Why do we need that, since we can always apply a UDF to a column during project, filter or join?

Indeed, a derived column consumes O(N) storage, where N is the number of rows in the table. We still need them because, the performance benefits outweigh the disadvantage of extra storage it consumes. Let us understand with the following use case example:

A compute engine like Presto can easily push down a filter predicate e.g. SELECT col1, col2, FROM table T1 WHERE col1='constant_value' , this allows for pruning the number of rows required for TableScan by applying the filtering WHERE col1=’constant_value’. This is not true of when a UDF is involved in the filter predicate, let us take an example SELECT col1, col2, FROM table T1 WHERE lower(col1)='constant_value'. While optimizers can easily push down the filter predicate, however, it can not be used in filtering using the lower and upper bound metrics, for example Iceberg manifest statics and Parquet row group statistics. As a result, we end up scanning a large number of rows.

So, to support push down of certain predicates (with UDFs in them) and reduce the amount of data scanned, derived column bring massive performance improvements. Derived columns have already been proven in RDBMS system e.g. DB2 [1], and now we intend to bring them to Presto.

### DIFF
--- a/RFC-0025-derived-column-support.md
+++ b/RFC-0025-derived-column-support.md
@@ -1,0 +1,575 @@
+# RFC-0025 Derived column
+
+## Proposers
+
+* Prashant Sharma
+* Tim Meehan
+
+## Related Issues
+
+* https://github.com/prestodb/presto/issues/27436
+
+## Summary
+
+Derived column is an existing feature for traditional database systems [1], it is a new concept for modern distributed database system. This RFC will cover what it takes
+to bring derived columns support to presto and the benefits.
+
+**What is a derived column?**
+
+A column created by applying a SQL expression or a UDF to one or more existing columns in a table.
+
+**Why do we need that, since we can always apply a UDF to a column during project, filter or join?**
+
+Indeed, a derived column consumes O(N) storage, where N is the number of rows in the table. We still need them because, the performance benefits outweigh the disadvantage
+of extra storage it consumes. Let us understand with the following use case example:
+
+A compute engine like Presto can easily push down a filter predicate e.g. `SELECT col1, col2, FROM table T1 WHERE col1='constant_value'` , this allows for pruning the number of
+rows required for TableScan by applying the filtering WHERE col1=’constant_value’. This is not true of when a SQL expression is involved in the filter predicate, let us take an
+example `SELECT col1, col2, FROM table T1 WHERE lower(col1)='constant_value'`. While optimizers can easily push down the filter predicate, however, it can not be used
+in filtering using the lower and upper bound metrics, for example Iceberg manifest statics and Parquet row group statistics. As a result, we end up scanning a large number of rows.
+
+So, to support push down of certain predicates (non-monotonic) and reduce the amount of data scanned, derived column bring massive performance improvements.
+Derived columns have already been proven in RDBMS system e.g. DB2 [1], and now we intend to bring them to Presto.
+
+### Problem
+
+Let us take the following example:
+
+```
+presto:test> explain select *from taxis where lower(store_and_fwd_flag) != lower('N');
+                                                                                                                                                                                 Query Plan                                                                                                                                                                                 
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ - Output[PlanNodeId 6][vendor_id, trip_id, trip_distance, fare_amount, store_and_fwd_flag] => [vendor_id:bigint, trip_id:bigint, trip_distance:decimal(5,2), fare_amount:double, store_and_fwd_flag:varchar]                                                                                                                                                               
+         Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: 500.00, memory: 0.00, network: ?}                                                                                                                                                                                                                                                                       
+     - RemoteStreamingExchange[PlanNodeId 207][GATHER - COLUMNAR] => [vendor_id:bigint, trip_id:bigint, trip_distance:decimal(5,2), fare_amount:double, store_and_fwd_flag:varchar]                                                                                                                                                                                         
+             Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: 500.00, memory: 0.00, network: ?}                                                                                                                                                                                                                                                                   
+         - ScanFilter[PlanNodeId 0,236][table = TableHandle {connectorId='iceberg2', connectorHandle='taxis$data@2530820414725702192', layout='Optional[taxis$data@2530820414725702192]'}, filterPredicate = (lower(store_and_fwd_flag)) <> (VARCHAR'n')] => [vendor_id:bigint, trip_id:bigint, trip_distance:decimal(5,2), fare_amount:double, store_and_fwd_flag:varchar] 
+                 Estimates: {source: CostBasedSourceInfo, rows: 6 (250B), cpu: 250.00, memory: 0.00, network: 0.00}/{source: CostBasedSourceInfo, rows: ? (?), cpu: 500.00, memory: 0.00, network: 0.00}                                                                                                                                                                    
+                 vendor_id := 1:vendor_id:bigint (1:22)                                                                                                                                                                                                                                                                                                                     
+                 trip_id := 2:trip_id:bigint (1:22)                                                                                                                                                                                                                                                                                                                         
+                 fare_amount := 4:fare_amount:double (1:22)                                                                                                                                                                                                                                                                                                                 
+                 trip_distance := 3:trip_distance:decimal(5,2) (1:22)                                                                                                                                                                                                                                                                                                       
+                 store_and_fwd_flag := 5:store_and_fwd_flag:varchar (1:22)                                                                                                                                                                                                                                                                                                  
+                                                                                                                                                                                                                                                                                                                                                                            
+(1 row)
+
+Query 20260326_110542_00004_bmix8, FINISHED, 1 node
+http://127.0.0.1:8080/ui/query.html?20260326_110542_00004_bmix8
+Splits: 1 total, 1 done (100.00%)
+[Latency: client-side: 126ms, server-side: 115ms] [0 rows, 0B] [0 rows/s, 0B/s]
+```
+
+1. `lower('N')` is constant folded and pushed down along with the predicate `filterPredicate = (lower(store_and_fwd_flag)) <> (VARCHAR'n')` to the Table scan node. However,
+   we see `lower(store_and_fwd_flag)` stays as is, table scan cannot perform this filtering efficiently as the underlying connector does not know about the UDF lower,
+   and thus ends up scanning large amount of data. The reason is, entire data is first brought inside presto and then filtering is performed.
+2. While reading from Table formats e.g. Iceberg, this problem take another shape all together. In Iceberg table, we can easily push down the filter to table scan node,
+   but we cannot use it to decide which files to skip for reading. Let me explain this in more detail, Iceberg manifest files store lower and upper bound metrics for each
+   column and these metrics decide whether to skip the file for reading or include it. This also happens at file format level, for example Parquet footer statics have the
+   information which row group should be having that information. In the case of `lower(col)` UDF function, the sorting order of the column is modified and as a result we cannot
+   simply
+   compute `lower(lower_bound_of_column)` and `upper(upper_bound_of_column)` to get the same file skipping behavior. This is because `lower` is not a monotonic function.
+3. We see `?` in cost estimates, because estimator does not know anything about the UDF `lower` it is not able to perform cost estimates. As a result, in a more complex
+   query, when such UDF's appear the join order can be a poor choice, and Plan can be a suboptimal plan. This was also addressed in [RFC-0005](RFC-0005-functions-stats.md).
+   In this approach the missing estimates for UDF's can also be addressed by replacing them with derived columns.
+
+### How does derived columns help?
+
+1. `lower(store_and_fwd_flag)` becomes `_1_store_and_fwd_flag`, then we can rewrite the query as `select *from taxis where _1_store_and_fwd_flag != lower('N');` This query
+   can be executed as a regular query, and the problem of underlying system unable to prune data for scanning is solved, because now we are dealing with columns instead of UDFs.
+2. For the case of Iceberg Table Format, it stores metrics and stats per column (and even bloom filters) in it's metadata that helps skip data files from scanning.
+3. The newly derived columns `_1_store_and_fwd_flag` has the stats on it, this gives better selectivity estimates and as a result optimizers can do a better job at
+   planning and coming up with an optimal plan. Therefore, problem of missing stats is solved as well.
+
+### Other Computing Engines
+
+#### Spark and Trino
+
+Spark and Trino currently does not do any special handling for this, i.e. Derived column is a novel concept to all distributed database engines.
+
+This spec does not focus on what changes are needed to support this feature in other Distributed database Engines than Prestodb.
+
+#### Mysql, MS-SQL, IBM DB2, Maria DB.
+
+All of the above support this feature as per the defined SQL standard.
+
+1. Maria DB: https://mariadb.com/docs/server/reference/sql-statements/data-definition/create/generated-columns
+2. IBM DB2: https://www.ibm.com/docs/en/db2-for-zos/12.0.0?topic=statements-alter-table#db2z_sql_altertable__syntaxdiagram_app_ztw_21c__title__1
+3. MS-SQL: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-ver17
+
+## Goals
+
+* Align with SQL Standard for Derived column followed by many RDBMS systems. [Ref](#mysql-ms-sql-ibm-db2-maria-db)
+* Present the derived column as a possible solution for solving the data skipping at the TableScan operator.
+* Propose metadata changes to Iceberg table properties to provide a portable solution for supporting Derived column .
+* Cover the necessary planner changes to benefit from Derived columns.
+* Discuss performance benchmarks of with and without this feature.
+* Remain backward compatible with previous versions of presto engines and other engines who have not yet adopted derived columns.
+
+The ultimate goal of all of the above is that we can allow data skipping at metadata level and derive performance benefit at the expense of some extra storage.
+
+## Proposed Implementation
+
+In this design, we will focus on Iceberg format initially and lay foundation for extending to other connectors.
+
+### Presto DDL changes:
+
+Create Table:
+
+```sql
+CREATE TABLE [ IF NOT EXISTS ]
+table_name (
+  { column_name data_type [GENERATED ALWAYS]  AS   ( <expression> ) [VIRTUAL | PERSISTENT] [NOT NULL] [ COMMENT comment ] [ WITH ( property_name = expression [, ...] ) ]
+  | LIKE existing_table_name [ { INCLUDING | EXCLUDING } PROPERTIES ]
+  | [ CONSTRAINT constraint_name ] { PRIMARY KEY | UNIQUE } ( { column_name [, ...] } ) [ { ENABLED | DISABLED } ] [ [ NOT ] RELY ] [ [ NOT ] ENFORCED ] }
+  [, ...]
+)
+[ COMMENT table_comment ]
+[ WITH ( property_name = expression [, ...] ) ]
+```
+
+For example:
+
+```sql
+CREATE TABLE test_table1 (                     
+    "c1" bigint,                                                 
+    "c2" varchar,                                                
+    "c3" double,
+    "c2_derived" varchar AS lower(c2) PERSISTENT
+ )                                                               
+
+```
+
+Alter Table:
+
+```sql 
+ALTER TABLE [ IF EXISTS ] name ADD COLUMN [ IF NOT EXISTS ] column_name data_type [GENERATED ALWAYS]  AS   ( <expression> ) [VIRTUAL | PERSISTENT]
+``` 
+
+GENERATED ALWAYS: Enforce that the derived column, cannot be directly insert via INSERT statements.
+
+VIRTUAL: Create a derived column, however do not persist it's data. It is always generated using the specified derived column SQL Expression for the column.
+Insert to a VIRTUAL columns are not permitted. The current version of RFC does not cover implementation of VIRTUAL columns and is mentioned in Future Work.
+
+PERSISTENT: Create a derived column and also persist it's data by generating using the specified derived column SQL Expression.
+
+### Caveats: Deviations from SQL standard.
+
+1. We allow direct insert into derived columns. If the table is written by another engine which does not support derived columns, write to derived columns cannot be restricted.
+2. It is thus responsibility of the user to maintain derived column information as up to date. Any stale entry in derived column will lead optimizer rewrite to produce incorrect
+   result when the derived column feature flags are enabled.
+3. We can provide warning, the derived column data is out of sync. And then it is user's responsibility they can drop and recreate the column. (Future work).
+4. In addition to the above, there are other possibilities for example the builtin UDF itself is updated by the plugin owner - such events are very hard to auto-detect.
+   Since the builtin UDFs are not versioned it can lead to possibly incorrect results. 
+5. We do not support UDFs with 'side effects', however, since there are no ways to detect what the UDF may be doing - it is hard to detect.
+
+### What is allowed in derived column expressions
+
+1. All built-in deterministic scalar functions.
+2. AND / OR expressions.
+3. Arithmetic expressions both binary and unary.
+4. If and CASE Statements.
+5. isNull and Cast expressions.
+6. Deterministic UDFs
+
+### What is not allowed in derived columns expressions
+
+1. Any of the aggregate functions.
+2. A sub query expressions.
+3. IN Query and Exists query.
+4. Non-deterministic functions.
+
+### Table properties
+
+Following table properties are added.
+
+1. `presto.derived-columns.spec.json` :
+    ```json 
+   {
+            "derivedColumnSpecs" : [ {
+               "derivedColumnType" : "PERSISTENT",
+               "derivedColumnExpression" : "SQL expression",
+               "derivedColumnName" : "derived Column"
+               }, {
+               "derivedColumnType" : "VIRTUAL",
+               "derivedColumnExpression" : "SQL expression2",
+               "derivedColumnName" : "derived Column2"
+               }]
+    }
+    ```
+3. This table property is hidden and not intended for the user to be entered/updated manually,
+   rather they are autopopulated when user performs the CREATE TABLE or ALTER TABLE query.
+4. There properties are persisted inside the table properties of the iceberg table i.e. metadata.json.
+   for example:
+
+```json
+{
+  "properties": {
+    "presto.derived-columns.spec.json": "{\n  \"derivedColumnSpecs\" : [ {\n    \"derivedColumnType\" : \"PERSISTENT\",\n    \"derivedColumnExpression\" : \"lower(c2)\",\n    \"derivedColumnName\" : \"c2_derived\"\n  }, {\n    \"derivedColumnType\" : \"PERSISTENT\",\n    \"derivedColumnExpression\" : \"c1*10.5\",\n    \"derivedColumnName\" : \"c1_derived\"\n  } ]\n}",
+    "write.format.default": "PARQUET",
+    "write.delete.mode": "merge-on-read",
+    "write.metadata.previous-versions-max": "100",
+    "write.parquet.compression-codec": "ZSTD",
+    "write.metadata.metrics.max-inferred-column-defaults": "100",
+    "write.metadata.delete-after-commit.enabled": "false",
+    "commit.retry.num-retries": "4",
+    "write.update.mode": "merge-on-read",
+    "read.split.target-size": "134217728"
+  }
+}
+```
+
+### Feature Flags
+
+1. `derived_columns.enable` : Boolean feature flag to enable or disable derived column feature for query rewrite.
+
+### Session Flag
+
+1. `derived_columns_enabled` : Boolean session flag to enable or disable derived column feature for query rewrite per session.
+
+### SPI changes to ColumnMetadata,
+
+```java
+/**
+ * This class stores the derived column information.
+ */
+public class DerivedColumnSpec
+{
+    private final DerivedColumnType derivedColumnType;
+    private final String derivedColumnExpression;
+    private final String derivedColumnName;
+    private final int derivedColumnFieldId;
+    private final String derivedColumnReturnType;
+
+    /**
+     * The field derivedColumnType, derivedColumnFieldId and derivedColumnReturnType are used for validation only,
+     * these values establish if derived column information has gone stale (by an external update) and needs a refresh.
+     *
+     * @param derivedColumnType A derived column can either be a GENERATED ALWAYS and PERSISTENT or VIRTUAL or just PERSISTENT.
+     * @param derivedColumnExpression A derived column expression, a generic SQL expression that presto recognizes.
+     * @param derivedColumnName Name of the derived column
+     * @param derivedColumnFieldId field ID is connector dependent sequence number for the column.
+     * @param derivedColumnReturnType return type of this column, used for validation purpose.
+     */
+    @JsonCreator
+    public DerivedColumnSpec(
+            @JsonProperty("derivedColumnType") DerivedColumnType derivedColumnType,
+            @JsonProperty("derivedColumnExpression") String derivedColumnExpression,
+            @JsonProperty("derivedColumnName") String derivedColumnName,
+            @JsonProperty("derivedColumnFieldId") Integer derivedColumnFieldId,
+            @JsonProperty("derivedColumnReturnType") String derivedColumnReturnType)
+    {
+        this.derivedColumnType = requireNonNull(derivedColumnType, "derivedColumnType is null");
+        this.derivedColumnExpression = requireNonNull(derivedColumnExpression, "derivedColumnExpression is null");
+        this.derivedColumnName = requireNonNull(derivedColumnName, "derivedColumnName is null");
+        this.derivedColumnFieldId = requireNonNull(derivedColumnFieldId, "derivedColumnFieldId is null");
+        this.derivedColumnReturnType = requireNonNull(derivedColumnReturnType, "derivedColumnReturnType is null");
+    }
+}
+```
+
+ColumnMetadata is enhanced with a new optional field of type `DerivedColumnSpec`.
+
+```
+   private final Optional<DerivedColumnSpec> derivedColumnSpec;
+
+    private ColumnMetadata(
+            String name,
+            Type type,
+            boolean nullable,
+            String comment,
+            String extraInfo,
+            boolean hidden,
+            Map<String, Object> properties,
+            Optional<DerivedColumnSpec> derivedColumnSpec)
+    {
+        checkNotEmpty(name, "name");
+        requireNonNull(type, "type is null");
+        requireNonNull(properties, "properties is null");
+
+        this.name = name;
+        this.type = type;
+        this.nullable = nullable;
+        this.comment = comment;
+        this.extraInfo = extraInfo;
+        this.hidden = hidden;
+        this.properties = properties.isEmpty() ? emptyMap() : unmodifiableMap(new LinkedHashMap<>(properties));
+        this.derivedColumnSpec = derivedColumnSpec;
+    }
+
+```
+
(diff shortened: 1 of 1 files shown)


RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,400 @@
...
+
+1. `derived-columns` : List of columns that are derived columns. 
+2. `derived-columns.spec.udf.json`: Full JSON specification for each derived column :
+
+```json
+{
+  "udfSpecList": [
+    {
+      "functionName": "_UDF_1",
+      "params": [
```
aditi-pandit: Why is params separated from arguments ? Are you using any special Iceberg or Presto UDF format here ?

ScrapCodes: @aditi-pandit , the way we can identify a UDF in presto is:
`catalog.schema.function_name` and it's params type. 

Once UDF is found, next step in checking it for replacement is, it's arguments should match too. Then we also need the arguments to match with !

Also, once Iceberg UDF [spec](https://iceberg.apache.org/udf-spec/#overview) is implemented, we should be able to uniquely identify and load a UDF by same search i.e. function name (id) and params. 

Interestingly, their entire UDF metadata spec does not store a function name :)

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,394 @@
...
+
+1. `derived_columns.enable` : Boolean flag to enable or disable derived column feature at global level.
+2. `derived_columns.plan_rewrite.enable_select_queries`: Boolean flag for enabling plan rewrite for select queries.
+3. `derived_columns.plan_rewrite.enable_insert_queries`: Boolean flag for enabling plan rewrite for insert queries.
+
+### Plan node changes for SELECT's filter
+
+1. Load the UDF from UDF spec provided in table properties.
+2. Scan the filter predicate and look for the UDF
+3. If the UDF defined in the spec is found, rewrite the filter predicate by replacing the UDF call with derived column.
```
jja725: feel like we can also rewrite the projection

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,394 @@
...
+   queries themselves are modified. This is also not desirable.
+
+#### Apache Iceberg. (TBD: in progress at https://github.com/apache/iceberg/issues/15923)
+Currently, we are using table properties for storing the required metadata for supporting derived column feature. It provides the required portability, but limited by
+certain aspects. For example, adding a derived column currently requires two steps i.e. adding a column manually and setting derived column configuration on it. 
+If there is a broader acceptance of this feature in the community, then we can have a first class support for derived columns for example:
+
+Alter table syntax
+```sql
+ALTER TABLE table_name {ADD | ALTER} COLUMN column_name data_type [DERIVED {[udf("column_name1")] | [SQL EXPRESSION]}]
```
jja725: do we need versioning for UDF?

jja725 (commented):
Agree that how write work would be the main concern here with compatibility with all the engine

RFC-0025-derived-column-support.md
ScrapCodes: Yes, we can rewrite projections too. However, the performance gain is negligible in most cases. There may be some cases, where this would be helpful e.g. UDF itself does a very compute intensive operation. 
Will try to cover this, thanks for suggesting!

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,394 @@
...
+
+**Why do we need that, since we can always apply a UDF to a column during project, filter or join?**
+
+Indeed, a derived column consumes O(N) storage, where N is the number of rows in the table. We still need them because, the performance benefits outweigh the disadvantage
+of extra storage it consumes. Let us understand with the following use case example:
+
+A compute engine like Presto can easily push down a filter predicate e.g. `SELECT col1, col2, FROM table T1 WHERE col1='constant_value'` , this allows for pruning the number of
+rows required for TableScan by applying the filtering WHERE col1=’constant_value’. This is not true of when a UDF is involved in the filter predicate, let us take an
+example `SELECT col1, col2, FROM table T1 WHERE lower(col1)='constant_value'`. While optimizers can easily push down the filter predicate, however, it can not be used
+in filtering at TableScan, simply because the underlying system is unaware of the UDF lower. As a result, we end up scanning a large number of rows.
```
tdcmeehan: More accurately, while we will push down the lower into the table scan and filter at the row level, as we do any other expression, what we can't do is use this expression with range (min/max) index values.  Some examples include Iceberg manifest statistics and Parquet rowgroup statistics.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,394 @@
...
+(1 row)
+
+Query 20260326_110542_00004_bmix8, FINISHED, 1 node
+http://127.0.0.1:8080/ui/query.html?20260326_110542_00004_bmix8
+Splits: 1 total, 1 done (100.00%)
+[Latency: client-side: 126ms, server-side: 115ms] [0 rows, 0B] [0 rows/s, 0B/s]
+```
+
+1. `lower('N')` is constant folded and pushed down along with the predicate `filterPredicate = (lower(store_and_fwd_flag)) <> (VARCHAR'n')` to the Table scan node, but
+   table scan cannot perform this filtering as the underlying connector does not know about the UDF lower or simply does not have the ability to evaluate it on lower and upper bounds.
```
tdcmeehan: The UDF can be evaluated, what can't be evaluated is this expression over lower and upper bounds, since those statistics are calculated on the base values.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,394 @@
...
+   In a jdbc based connector the underlying connector may be able to evaluate the udf if there exist a udf with same name in the underlying database system. Since we do not know,
```
tdcmeehan: I believe this is orthogonal to JDBC connector function pushdown.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,413 @@
...
+```json
+{
+  "properties": {
+    "write.format.default": "PARQUET",
+    "write.delete.mode": "merge-on-read",
+    "derived-columns": "c2_derived",
+    "write.metadata.previous-versions-max": "100",
+    "write.parquet.compression-codec": "ZSTD",
+    "write.metadata.metrics.max-inferred-column-defaults": "100",
+    "derived-columns.spec.udf.json": "{\n  \"udfSpecList\" : [ {\n    \"catalog\" : \"presto\",\n    \"schema\" : \"default\",\n    \"functionName\" : \"lower\",\n    \"params\" : [ \"varchar\" ],\n    \"arguments\" : [ {\n      \"argumentIndex\" : 0,\n      \"argumentType\" : \"varchar\",\n      \"argumentValue\" : \"c2\",\n      \"columnRef\" : \"COLUMN\"\n    } ],\n    \"derivedColumnName\" : \"c2_derived\",\n    \"returnType\" : \"varchar\"\n  } ]\n}",
```
tdcmeehan:
```suggestion
    "presto.derived-columns.spec.udf.json": "{\n  \"udfSpecList\" : [ {\n    \"catalog\" : \"presto\",\n    \"schema\" : \"default\",\n    \"functionName\" : \"lower\",\n    \"params\" : [ \"varchar\" ],\n    \"arguments\" : [ {\n      \"argumentIndex\" : 0,\n      \"argumentType\" : \"varchar\",\n      \"argumentValue\" : \"c2\",\n      \"columnRef\" : \"COLUMN\"\n    } ],\n    \"derivedColumnName\" : \"c2_derived\",\n    \"returnType\" : \"varchar\"\n  } ]\n}",
```

ScrapCodes: One of the goal is, that this feature is portable and thus we would expect other engines to be able to respect this table property eventually. So, I will not prefix `presto`

tdcmeehan: Portability won't be possible because the expressions will have different semantics between engines.  Just think about all of the minute differences in function behavior between Presto and Spark functions.  It's kind of like a more focused version of the view definition compatibility challenge, which the Iceberg spec itself punts on by specifying the dialect along with the view definition.  We should take the same tack here: if an engine really wants to attempt to interpret the derived column expressions, it must commit to _Presto's dialect_.

Attempting to solve cross-engine compatibility in this effort seems too large, and I would defer any such energy to the Iceberg spec itself, where eventually we'll have UDF support and potentially some subset of Substrait as a common definition medium between engines.  IMO it shouldn't be in scope of this effort.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,413 @@
...
+               ->     "c1" bigint,                                                 
+               ->     "c2" varchar,                                                
+               ->     "c3" double,
+               ->     "c2_derived" varchar
+               ->  )                                                               
+               ->  WITH (                                                          
+               ->     "format-version" = '3',
+               ->     location = 's3://test-bucket/perf_test/test_table9',
+               ->     "derived-columns" = Array['c2_derived'],
+               ->     "derived-columns.spec.udf.json" = JSON '{
```
tdcmeehan: I think the granularity is wrong here.  Right now we're scoping this to UDFs.  UDFs are functions, and what's implicit is that this should be a _function _call__.  A function call is just one type of _expression_.  I think what should be stored here instead is an _expression_, because that would capture the use cases better.  For example, the derived column might be an IF statement, which is not a UDF.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,413 @@
...
+
+1. `lower(store_and_fwd_flag)` becomes `_1_store_and_fwd_flag`, then we can rewrite the query as `select *from taxis where _1_store_and_fwd_flag != lower('N');` This query
+   can be executed as a regular query, and the problem of underlying system unable to prune data for scanning is solved. For example, Iceberg stores metrics and stats per column
+   in it's metadata that helps skip data files from scanning.
+2. The newly derived columns `_1_store_and_fwd_flag` has the stats on it, this gives better selectivity estimates and as a result optimizers can do a better job at
+   planning and coming up with an optimal plan.
+
+### Other Computing Engines
+
+#### Spark and Trino
```
tdcmeehan, line 85: Many relational databases support this feature.  We should refer to them.  For example from MariaDB: https://mariadb.com/docs/server/reference/sql-statements/data-definition/create/generated-columns

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,413 @@
...
+
+In this design, we will focus on Iceberg format initially and lay foundation for extending to other connectors.
+
+### Table properties 
+
+Following table properties are added.
+
+1. `derived-columns` : List of columns that are derived columns. 
+2. `derived-columns.spec.udf.json`: Full JSON specification for each derived column :
+
```
tdcmeehan: We do not want the user to manually specify a JSON blob to use the derived expression.  Here is how it should work, very high level:

1. We add parser support for GENERATED ALWAYS, see example here: http://mariadb.com/docs/server/reference/sql-statements/data-definition/create/generated-columns#syntax

2. Parse and analyze the expression
3. During createtable, we store metadata (raw SQL expression) for the derived column to be persisted by the connector
4. The connector returns the metadata when it retrieves the table definition
5. This is analyzed by the query which queries the table, and there is an optimizer rule to leverage it where appropriate (very simple, exact match of expression)

RFC-0025-derived-column-support.md
ScrapCodes: Great point !

I have interpretted this in following two possible ways 

1. We store expression as string and do a string match for derived column replacement !
2. We store expression as expression tree of some sort and compute a canonical form of user queries expressions and then try to match them.

While, I think you are alluding to first point more than second, the result of both is same. User's expressions get swapped when they are "exactly same". In later approach, the chance of semantically checking and even if user's expression occurs as sub expression - achieving a replacement is possible.

Let's review the UDF approach again, 

1. Can we create a UDF out of any user provided expression ? If yes, then we can have a UDF with just IF statement and setup derived column.
2. We can achieve rewrite even if UDF occurs as a subexpression, for example : if we have a derived column setup on `lower(c2)`, but in our query we have: ` explain SELECT c1, c2 FROM test_table2 WHERE (concat('A', lower(c2)) ) = 'Aa';` we can still achieve correct rewrite.

```
presto:perf_test> explain SELECT c1, c2 FROM test_table2 WHERE (concat('A', lower(c2)) ) = 'Aa';
                                                                                                                                                       Query Plan                                                                                                                                                        
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 - Output[PlanNodeId 6][c1, c2] => [c1:bigint, c2:varchar]                                                                                                                                                                                                                                                               
         Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: 714.00, memory: 0.00, network: ?}                                                                                                                                                                                                                    
     - RemoteStreamingExchange[PlanNodeId 219][GATHER - COLUMNAR] => [c1:bigint, c2:varchar]                                                                                                                                                                                                                             
             Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: 714.00, memory: 0.00, network: ?}                                                                                                                                                                                                                
         - ScanFilter[PlanNodeId 0,252][table = TableHandle {connectorId='iceberg', connectorHandle='test_table2$data@7072973623560046354', layout='Optional[test_table2$data@7072973623560046354]'}, filterPredicate = (concat(VARCHAR'A', c2_derived)) = (VARCHAR'Aa')] => [c2:varchar, c2_derived:varchar, c1:bigint] 
                 Estimates: {source: CostBasedSourceInfo, rows: 3 (357B), cpu: 357.00, memory: 0.00, network: 0.00}/{source: CostBasedSourceInfo, rows: ? (?), cpu: 714.00, memory: 0.00, network: 0.00}                                                                                                                 
                 c2 := 2:c2:varchar (1:28)                                                                                                                                                                                                                                                                               
                 c2_derived := 4:c2_derived:varchar                                                                                                                                                                                                                                                                      
                 c1 := 1:c1:bigint (1:28)                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                         
(1 row)

Query 20260511_125853_00003_yt5q2, FINISHED, 1 node
http://127.0.0.1:8080/ui/query.html?20260511_125853_00003_yt5q2
Splits: 1 total, 1 done (100.00%)
[Latency: client-side: 75ms, server-side: 65ms] [0 rows, 0B] [0 rows/s, 0B/s]
```

Also in string matching, a user's query may not benefit - even if there are differences in spaces while writing the expression. While it's fair to say that is user error, the udf approach currently proposed does not suffer from it.

The advantages of expression and string matching are following :
1. We are consistent with rdbms ! 
2. It's easy to implement

If consistency with RDBMS is strict requirement, then we should definitely support expressions. A full fledged expression matching will be much more elaborate effort to implement and risk of missing a canonical form is still higher than UDF approach, and if we go with string matching we lose out on many possible matches due to just differences in naming/canonical form/spaces.

tdcmeehan: @ScrapCodes I don't mean we should swap in full text strings.  What we can do is treat derived column expressions as views, except scoped at the expression level, whereas views are scoped at the query level.  We store the raw text in the connector, same as we do for views.  During analysis, we parse the text and this feeds into the planner, which then will use this to create projections where appropriate when the plan is created.  It won't be full text matched, it will be matched either at the AST level, or at the plan level (my preference is plan level).

The most compelling reason to go this route is that it's simply how it's defined in the SQL standard, and we typically defer to the standard when it weighs in on something.

```
<select sublist>   ::= <derived column> | <qualified asterisk>
<derived column>   ::= <value expression> [ <as clause> ]
<as clause>        ::= [ AS ] <column name>
```

RFC-0025-derived-column-support.md
ScrapCodes: @tdcmeehan , we will still need a mapping between "user's expression" -> "derived_column_name". Plain strings cannot hold this info for a table with more than one derived column.

RFC-0025-derived-column-support.md
ScrapCodes: My preference is also matching via generating Plan, however it will require us to write a query with SELECT and FILTER/PROJECT and then convert it into plan node. (running as the current user's permissions). Because, we cannot simply run/analyze/plan a headless expression. 

Once the Plan is generated we can extract the plan node for expression and try to do matching.

Also, we will need to account additional time required to compute plan which will contribute to the overall query running time.

RFC-0025-derived-column-support.md
tdcmeehan: Yes--the Iceberg connector can create and manage this mapping.  It need not be user-managed.

RFC-0025-derived-column-support.md
tdcmeehan: Agreed, but the planning time shouldn't be an issue.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,433 @@
...
+    "c3" double,
+    "c2_derived" varchar AS lower(c2) PERSISTENT
+ )                                                               
+
+```
+
+Alter Table:
+
+```sql 
+ALTER TABLE [ IF EXISTS ] name ADD COLUMN [ IF NOT EXISTS ] column_name data_type [GENERATED ALWAYS]  AS   ( <expression> ) [VIRTUAL | PERSISTENT]
```
tdcmeehan, line 145: Should we always only allow `PERSISTENT` at first, as well as `GENERATED ALAYS`?

ScrapCodes: We can allow PERSISTENT at first, my thought was if we keep only PERSISTENT then it becomes redundant terminal node during parsing. So, let me know how you think about it, can alter either way. My preference is to keep both VIRTUAL and PERSISTENT, and then later throw an error if VIRTUAL is used - "current connector does not support VIRTUAL for derived columns."

RFC-0025-derived-column-support.md
ScrapCodes: My preference was to go with just UDF, there is a merit in supporting expression for the convenience they bring. However, they lack portability. UDFs on the other hand are portable - Iceberg has support to store a UDF in the table format itself. But this is an endless debate, and either way we can bring portability. 

For example, it is always possible to wrap an expression in a UDF and a UDF is already an Expression.

RFC-0025-derived-column-support.md
ScrapCodes: Since we moved to Expression based syntax, this is not applicable. If the UDF is versioned, I guess we will just be picking the default version.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+
+A column created by applying a SQL expression or a UDF to an existing column in a table.
+
+**Why do we need that, since we can always apply a UDF to a column during project, filter or join?**
+
+Indeed, a derived column consumes O(N) storage, where N is the number of rows in the table. We still need them because, the performance benefits outweigh the disadvantage
+of extra storage it consumes. Let us understand with the following use case example:
+
+A compute engine like Presto can easily push down a filter predicate e.g. `SELECT col1, col2, FROM table T1 WHERE col1='constant_value'` , this allows for pruning the number of
+rows required for TableScan by applying the filtering WHERE col1=’constant_value’. This is not true of when a UDF is involved in the filter predicate, let us take an
```
aditi-pandit: Nit : Its not necessarily "UDF" but any SQL expression involving the columns of the table.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+``` 
+
+### Caveats: Deviations from SQL standard.
+
+1. We allow direct insert into derived columns. If the table is written by another engine which does not support derived columns, write to derived columns cannot be restricted.
+2. It is thus responsibility of the user to maintain derived column information as up to date. Any stale entry in derived column will lead optimizer rewrite to produce incorrect
+   result when the derived column feature flags are enabled.
+3. We can provide a command to sync derived columns when they do go out of sync. (Similar to Materialized views REFRESH).
+
+### What is allowed in derived column expressions
```
aditi-pandit, line 165: Maybe not a big deal but are IS NULL and NOT IS NULL allowed for derived columns ?

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+3. Arithmetic expressions both binary and unary.
+4. If and CASE Statements.
+
+### What is not allowed in derived columns expressions
+
+1. Any of the aggregate functions.
+2. A sub query expressions.
+3. IN Query and Exists query.
+
+### Table properties
```
aditi-pandit, line 181: Table properties are Iceberg specific. Do you have a design for Hive as well ?

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+    "c3" double,
+    "c2_derived" varchar AS lower(c2) PERSISTENT
+ )                                                               
+
+```
+
+Alter Table:
+
+```sql 
+ALTER TABLE [ IF EXISTS ] name ADD COLUMN [ IF NOT EXISTS ] column_name data_type [GENERATED ALWAYS]  AS   ( <expression> ) [VIRTUAL | PERSISTENT]
```
aditi-pandit, line 145: What is the difference between VIRTUAL and PERSISTENT ? Please can you elaborate.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+```
```
aditi-pandit, line 146: Do the derived column expressions show up in DESCRIBE TABLE/SHOW COLUMNS sql ? If yes, then what will be the syntax of the expressions.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+               "derivedColumnExpression" : "SQL expression2",
+               "derivedColumnName" : "derived Column2"
+               }]
+    }
+    ```
+3. These table properties are not intended for the user to be entered/updated manually, rather they are autopopulated when user enters the CREATE TABLE query or ALTER TABLE query.
+
+### Feature Flags
+
+1. `derived_columns.enable` : Boolean feature flag to enable or disable derived column feature for query rewrite.
```
aditi-pandit, line 223: Why are you proposing a flag for this ? Do you see any downsides to pick up derived columns by default ?

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+    "c2" varchar,                                                                                 
+    "c2_derived" varchar,                                                                         
+    "c4" double                                                                                   
+ )                                                                                                
+ WITH (                                                                                           
+    "derived-columns" = ARRAY['c2_derived'],                                                      
+    "derived-columns.spec.expression.json" = '{                                                   
+   "expressionSpecList" : [ {                                                                     
+     "derivedColumnType" : "PERSISTENT",                                                          
+     "derivedColumnExpression" : "if(lower(c2)=''c'',''g'',''e'')",
```
aditi-pandit: How do you handle the case if there is a type mismatch between what the user specifies as the type of the column and the expression return type ?

aditi-pandit: This seems like a very contrived expression. Can we use something simple like just "lower(column)" instead ?

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+
+### What is not allowed in derived columns expressions
+
+1. Any of the aggregate functions.
+2. A sub query expressions.
+3. IN Query and Exists query.
+
+### Table properties
+
+Following table properties are added.
```
aditi-pandit, line 183: These are in metadata.json file ?

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+
+1. `derived-columns` : List of columns that are derived columns.
+2. `derived-columns.spec.expression.json` :
+    ```json 
+   {
+            "udfSpecList" : [ {
+               "derivedColumnType" : "PERSISTENT",
+               "derivedColumnExpression" : "SQL expression",
```
aditi-pandit, line 190: Can you give more info about the SQL dialect of this expression ? Seems like you want atleast Presto and Spark to understand it.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+1) Suppose we append the extra column to the underlying data files and do not have schema evolution visible. Then append to the underlying files shall not corrupt it.
+2) What if there is addition of derived column and then addition of actual column and then again derived column and so on. This can be a problem, when a writer unaware of
+   derived column feature tries to write data. It can corrupt the internal records.
+3) Suppose we evolve the schema for Iceberg table on adding a derived column, this is the best alternative because it will not corrupt the data and we remain backward compatible
+   with engines unaware of this feature. However, it will break the requirement that insert from the engine unaware of the derived column feature will not work, unless the INSERT
+   queries themselves are modified. This is also not desirable.
+
+## Alternatives considered.
+
+### 1. Use table properties to configure derived column information (instead of alter column syntax).
```
aditi-pandit, line 440: Thought this is your actual solution. This writeup gives the impression its just an alternative considered. Please can you elaborate.

RFC-0025-derived-column-support.md
tdcmeehan: To be clear, deriving a common subset of expressions that are interpretable by both Spark and Presto is hard and likely outside of the scope of this RFC.  I think the most straightforward thing is to treat them like views, which defer on cross-platform interpretability and force any consumer of the view SQL to understand Presto's dialect.  Cross platform expressions can be considered an orthogonal yet important task.

RFC-0025-derived-column-support.md
ScrapCodes: Correct my understanding on this please, Do you mean a simple expression for example: c1 > 100 will be affected, I guess no. So it is always expressions involving UDFs (in presto's case a IF statement is also a UDF)?

RFC-0025-derived-column-support.md
ScrapCodes: I will add comments !

RFC-0025-derived-column-support.md
ScrapCodes: Will update this soon.

RFC-0025-derived-column-support.md
ScrapCodes: Yes

RFC-0025-derived-column-support.md
ScrapCodes: Good point, I have mentioned, the current design focuses only on Iceberg, and all other connectors are mentioned in Future Work.

It will be interesting work to support Hive, and may be that will be next phase of the RFC.

RFC-0025-derived-column-support.md
ScrapCodes: Yes, will include examples.

RFC-0025-derived-column-support.md
ScrapCodes: Yes, that point is covered under Caveats section. Basically, derived column data can easily go out of sync and then optimizer rewrite can render incorrect results. So just like Materialized view, it is user's responsibility to maintain a sync or disable using this feature flag.

RFC-0025-derived-column-support.md
ScrapCodes: if I am understanding correctly, Expression return type is interpreted so this situation should not occur.

ScrapCodes: I had a purpose in mind for this contrived expression, i.e. wanted to show the working of sub-expression matching. There is a usage guide section, where I have covered what users in general will see as an example.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,492 @@
...
+Splits: 0 total, 0 done (0.00%)
+[Latency: client-side: 48ms, server-side: 40ms] [0 rows, 0B] [0 rows/s, 0B/s]
+
+presto:perf_test> show create table test2;
+                                           Create Table                                           
+--------------------------------------------------------------------------------------------------
+ CREATE TABLE iceberg.perf_test.test2 (                                                           
+    "c1" bigint,                                                                                  
+    "c2" varchar,                                                                                 
+    "c2_derived" varchar,
```
ScrapCodes: Will soon have an update to showQueryRewrite with 
```
CREATE TABLE iceberg.perf_test.test2 (                                                           
    "c1" bigint,                                                                                  
    "c2" varchar,                                                                                 
    "c2_derived" varchar AS if(lower(c2) = 'c', 'g', 'e') PERSISTENT,                                                                         
```

RFC-0025-derived-column-support.md
ScrapCodes: This was my previous solution, where a user would manually enter the Table properties. Now that is no longer true, we support CREATE TABLE and ALTER TABLE syntax to achieve this. Internally table properties are used - but they are not end user editable.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,560 @@
...
+* https://github.com/prestodb/presto/issues/27436
+
+## Summary
+
+Derived column is an existing feature for traditional database systems [1], it is a new concept for modern distributed database system. This RFC will cover what it takes
+to bring derived columns support to presto and the benefits.
+
+**What is a derived column?**
+
+A column created by applying a SQL expression or a UDF to an existing column in a table.
```
hantangwangd: nit: should we use `one or more existing columns` here to be more precise?

RFC-0025-derived-column-support.md
hantangwangd: As I understand it, the supported expression types listed below seem broader than what would normally be considered UDFs:

```
1. All built-in scalar functions.
2. AND / OR expressions.
3. Arithmetic expressions both binary and unary.
4. If and CASE Statements.
5. isNull and Cast expressions.
```

Since UDF typically refers to User Defined Functions, it may be worth using a more general term here.

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,560 @@
...
+In this design, we will focus on Iceberg format initially and lay foundation for extending to other connectors.
+
+### Presto DDL changes:
+
+Create Table:
+
+```sql
+CREATE TABLE [ IF NOT EXISTS ]
+table_name (
+  { column_name data_type [GENERATED ALWAYS]  AS   ( <expression> ) [VIRTUAL | PERSISTENT] [NOT NULL] [ COMMENT comment ] [ WITH ( property_name = expression [, ...] ) ]
```
hantangwangd, line 121:
```suggestion
  { column_name data_type [[GENERATED ALWAYS]  AS   ( <expression> ) [VIRTUAL | PERSISTENT]] [NOT NULL] [ COMMENT comment ] [ WITH ( property_name = expression [, ...] ) ]
```

nit: missing a definition for normal column?

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,560 @@
...
+    "c3" double,
+    "c2_derived" varchar AS lower(c2) PERSISTENT
+ )                                                               
+
+```
+
+Alter Table:
+
+```sql 
+ALTER TABLE [ IF EXISTS ] name ADD COLUMN [ IF NOT EXISTS ] column_name data_type [GENERATED ALWAYS]  AS   ( <expression> ) [VIRTUAL | PERSISTENT]
```
hantangwangd, line 145:
```suggestion
ALTER TABLE [ IF EXISTS ] name ADD COLUMN [ IF NOT EXISTS ] column_name data_type [[GENERATED ALWAYS]  AS   ( <expression> ) [VIRTUAL | PERSISTENT]]
```

nit: missing a definition for normal column?

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,560 @@
...
+Insert to a VIRTUAL columns are not permitted. The current version of RFC does not cover implementation of VIRTUAL columns and is mentioned in Future Work.
+
+PERSISTENT: Create a derived column and also persist it's data by generating using the specified derived column SQL Expression.
+
+### Caveats: Deviations from SQL standard.
+
+1. We allow direct insert into derived columns. If the table is written by another engine which does not support derived columns, write to derived columns cannot be restricted.
+2. It is thus responsibility of the user to maintain derived column information as up to date. Any stale entry in derived column will lead optimizer rewrite to produce incorrect
+   result when the derived column feature flags are enabled.
+3. We can provide warning, the derived column data is out of sync. And then it is user's responsibility they can drop and recreate the column. (Future work).
```
hantangwangd, line 160: Just curious: how do we track and detect that a derived column is out of sync? And can we simply drop and re-add the derived column to sync it?

RFC-0025-derived-column-support.md
```diff
@@ -0,0 +1,560 @@
...
++
++    private ColumnMetadata(
++            String name,
++            Type type,
++            boolean nullable,
++            String comment,
++            String extraInfo,
++            boolean hidden,
++            Map<String, Object> properties,
++            DerivedColumnSpec derivedColumnSpec
```
hantangwangd: Looks like there's a missing `)` at the end. Also, is the diff message display here intended?

hantangwangd (commented):
Hi @ScrapCodes, thanks for this RFC. Regarding our eventual perspective on derived columns, I have some high-level questions and thoughts for discussion:

1. When adding a derived column to an existing table that already contains data, how are the values for existing rows handled?
2. When inserting rows, is the derived column: (a) not specifiable, (b) specifiable but auto-computed if omitted, or (c) mandatory (NULL when absent)?
3. For the tracking and detection of sync states of derived columns:
 - Is the engine responsible for tracking and detection? If so, how would that be implemented? (Considering that other unaware engines might perform writes, source columns could be updated, or even the function logic itself could change, etc.)
 - Or, the engine wouldn't manage the sync state—users are responsible for their own data. The engine would simply provide sync capability for users to call on demand, similar to `REFRESH MV` or `REWRITE_DATA_FILES`.
4. Filter and projection rewrite optimizations are indeed a huge performance advantage that derived columns can bring. But VIRTUAL-mode derived columns don't seem to offer that benefit—so what's the value of having them?

Ultimately, the core question is: in our final perspective, is a derived column a special type of column, or is it essentially more like a normal column with an optimization hint?

Looking at the overall design and the POC examples, it seems we're mainly considering the latter—where derived columns are essentially more like normal columns with an optimization hint. However, many of the concerns outlined in the "Considerations" and "Challenges" sections appear to stem from the former.

In my view, the gap between these two modes is quite significant. I think it would be helpful to first clearly define the boundaries—specifically, what is feasible and on our roadmap versus what is out of scope or not advisable.

Just some initial thoughts on my side—happy to be corrected if I've misunderstood anything, and looking forward to further discussion.

RFC-0025-derived-column-support.md
ScrapCodes: Hi @hantangwangd and @aditi-pandit , Thank you for the comment, I guess I understood now - there are certain expressions apart from just UDFs which can change the meaning of lower and upper bound used for file pruning (in future even bloom filters) and thus affect the ability to perform file pruning effectively.

RFC-0025-derived-column-support.md
ScrapCodes: Yes, a drop and re-add the derived column will trigger the expression to be applied on existing rows and new column updated.

In the absence of INSERT support for derived column we will have to do UPDATE e.g. `UPDATE test2 SET c2_derived = lower(c2)`.

hantangwangd (approved):
>Yes, the current RFC is going to focus on a) and b).
>
> My understanding is, derived column is opt in feature, just as a materialized view and it is the user's responsibility to ensure that derived columns data is in sync and then turn on the feature flag/session flag. In Presto, unlike any traditional database, we do not have the luxury of guarantee that data cannot be modified elsewhere and thus we cannot offer that guarantee either. However, what we can do is, as far as possible give hints of data possibly out of sync etc...

Totally agree! Looks good to me.

> Before we can start on c), d) and e), we will have to have answers for all the issues we have already discussed.

Thanks for the detailed example from MariaDb. Yes, I agree. It might be worth surveying a broad range of derived-column-supporting RDBMSs and thinking through the open issues we've talked about first, so that we're well positioned for c), d), and e).

RFC-0025-derived-column-support.md
hantangwangd: IMO we should be careful with this. Schema evolution probably shouldn't force data changes underneath — that's not really in line with modern lakehouse table format recommendations. What if we offered a dedicated COW-based sync operation (rather than MOR-based update statement), and left it to users to trigger when needed, like they do with `REFRESH MV` or `rewrite_data_files`?

ScrapCodes: Thanks, this is a good point - will keep this in mind.

aditi-pandit (approved):

